PR #46, opened 2015-02-06
fix inverted edit volumes placeholder

lucamartinetti:


vincent99:
d'oh, that is indeed backwards.  The square brackets would need to move too though, `[/path/on/host:]/path/in/container[:ro]`

lucamartinetti:
moved brackets too

vincent99:
Thanks!

PR #98, opened 2015-03-18
Move machine transitioning message, remove DigitalOcean 512mb

vincent99:


deniseschannon:
I had a conversation with @ibuildthecloud and he was saying that most people will want the 512mb for their hosts, so should we be removing that option from the DO page?

vincent99:
Compromising on leaving it as an option but defaulting to 1GB in rancherio/ui@7b585db

PR #139, opened 2015-04-30
Group order of DigitalOcean regions

deniseschannon:
@vincent99 Grouped the DO regions so that NYC 2/3 and Amsterdam 2/3 would be together.


vincent99:
They should really be sorted in the UI because this will break again next time I update the json, but this works for now..  #656

PR #185, opened 2015-06-03
Modified Add Host slider icon and other fine touchups

Nchikate:
@vincent99


vincent99:
These are incorporated into #202, except for the colors on the slider.

PR #204, opened 2015-06-11
Resolves pods not wrapping in service divs

stephenlawrence:
Fixes rancherio/rancher#1166


vincent99:
`.pods` is the container for the columns of hosts.  Removing `nowrap` from there will make the columns wrap on window resize and reflow while the JS layout adjusts how many columns should be shown and their widths.  I think you want something more like `.pod-info-line { white-space: initial }`

vincent99:
Changed in #207..

PR #217, opened 2015-06-20
Add href to host icons, possibly making them keyboard-focusable and screen reader-accessible

ndarilek:


vincent99:
Sorry it took so long to get to this, we've been swamped with DockerCon.

ndarilek:
Any idea when this will land in a release? Just tried a single-node install and it doesn't appear that these changes are there--that or something is very wrong.

vincent99:
We normally create a `0.WW-rc1` In DockerHub every Wednesday (`WW` being the week number of the year).  That gets looked at by our QA, some issues are maybe fixed with additional `-rcX` tags. When they're happy, usually Friday or Monday, the newest one gets promoted to `latest`.

This week I believe there was an update but we didn't change the UI tag it contains because I've been working on a large set of changes to update to Ember.js 1.13 and it wasn't stable yet.  So long story short, likely next Friday.

If you click on the Endpoint in the API & Keys page, then navigate up to `/v1/settings/api.ui.index`, that is the URL for the UI that the backend serves up. Clicking edit and replacing the version number with `latest` will always give you what's in the master branch of this repo, which would include your PR.   We use this internally and don't normally merge broken stuff to master but there's no strong guarantee that its stable at any given time.  (Hmm... Hopefully the API UI is accessible...)

ndarilek:
Where is the API and keys page? I've poked around the UI a bit but can't
seem to find it.

Thanks.

ndarilek:
Oh, looks like there may be more inaccessible UI. The Getting Started
page references an account icon which I can't find. Betting it's also
missing a href/alt/aria-label attribute.

PR #222, opened 2015-07-09
Add Exoscale host provider

vincentbernat:
Exoscale is a Swiss cloud provider and one of the provider currently
supported in Docker Machine. Its addition in Rancher UI follows a logic
similar to Amazon EC2 with a multi-step wizard:

 1. Ask for credentials.
 2. Grab the list of security groups.
 3. Ask which security group to use and propose to create the default
    one if it doesn't exist.
 4. Create the default security group if requested.
 5. Grab the instance profiles and available disk sizes.
 6. Let the user choose a name, instance profile and disk size.

Currently, only one availability zone is available. Therefore, there is
no drop-down list to choose one.

While there are several images available, we enforce the use of Ubuntu
Trusty. It would be possible to let the user choose, but this needs some
work to know which images would work and which wouldn't.

The default ingress rules also include the Docker daemon port. Docker
Machine needs it for the first time installation.

This PR needs https://github.com/rancherio/cattle/pull/667 to work as expected since Exoscale API only works over HTTPS. Also, `api.exoscale.ch` needs to be whitelisted. This can be done by setting the `CATTLE_API_PROXY_WHITELIST=api.exoscale.ch` environment variable. I will also do a PR for cattle for that.

If you want to test, you can register with the `DOCKERMACHINEROCKS` coupon code to get 20CHF of credit (the code works several times).


vincent99:
@vincentbernat This looks good.. I'll take a closer look and merge once rancher/cattle#667 and rancher/cattle#670 are merged so I can test it.

ibuildthecloud:
@vincent99 I've merged the cattle PRs.  Should land in latest in the morning.

vincent99:
Merged in #228 with some tweaks for the Ember 1.13 upgrade merged after you started;  Thanks @vincentbernat!

vincentbernat:
Great! And thanks for fixing the two bugs that I didn't notice!

PR #237, opened 2015-07-27
Hide Ubuntu 14.10

deniseschannon:
@vincent99 This should hide the Ubuntu 14.10 from DO. Assuming I found the right place. :)

(rancher/#1666)


deniseschannon:
@vincent99 15.04 isn't officially supported by Docker per the Docker docs so I wasn't sure if we want to include it.

PR #255, opened 2015-09-16
Added a better color for welcome screen “Learn more” link

jjperezaguinaga:
Noticed that the welcome screen has a hard to read link, due it inheriting the overall `link-color` mixin. I added a specific rule to make it look better just in the welcome screen, so hopefully it matches your standards.

screen shot 2015-09-17 at 01 35 53


vincent99:
Looks good to me..

PR #258, opened 2015-09-23
Ubiquity Hosting user interface

justacan:
Adding Ubiquity Hosting to Rancher UI.

Dependent pull requests include:
[Cattle](https://github.com/rancher/cattle/pull/886)
[Machine](https://github.com/rancher/machine/pull/8)


vincent99:
Hi Justin,

- Can you rebase off master, I just merged in [#259](https://github.com/rancher/ui/pull/259) with some changes that will be relevant to this.  Specifically
  - The `/proxy/` to hit APIs through cattle will now be under `/v1/proxy/`.
  - Use `Ember.Controller.extend` instead of `Ember.ObjectController.extend` (the PR converts all the other machine drivers)
- `allImages`, `defaultImage`, and similar properties should go on the controller and accessed without `model.`, not be put into the model.
  - Once you switch to `Ember.Controller`, use `model.` only for things that belong to the model returned from the route and that will be sent to the server/`save()`d.  `allImages` and similar can be accessed just like that, without `model.`.
- Pretty sure `btoa()` is in all the modern browsers, so you can just use that.
  - If there's some reason this is needed, please move it to `app/utils/base64.js` and have it use the native one when available.
- Can you please use e.g. `name: function() {...}.observes(...)` to match all the rest of the code (vs `name: Ember.observer(..., function() {...})`)
- Similarly, `returnsPromise.then(function() { /*success*/ }).catch(function() { /*fail*/})` vs `returnsPromise.then(function() { /*success*/ }, function() { /*fail*/})`
- The logo can be ~99% smaller if you save-as it without all the Illustrator metadata.

Then we can merge once the backend PRs are in.  Thanks

justacan:
We switched over to '/v1/proxy/' and now are getting the response:

{"id":"5d2203d9-2fc3-46b9-b43e-a06e2565857a","type":"error","links":{},"actions":{},"status":403,"code":"InvalidCSRFToken","message":"CSRF header and cookie do not match","detail":null}

We did not have this error in the past, how do we get past it?

vincent99:
The proxy path required a change in cattle too; try pulling to bring in https://github.com/rancher/cattle/pull/891 if using master, or use today's `v0.39.0-rc1`?  

The regular API has a CSRF check, where it sends a value in a cookie and checks that the request headers contain the same value in a header.  That *should* not be required for the proxy, but if you are hitting `/v1/proxy` on a version where it's still at `/proxy` the auth code would probably check for that before returning 404 for a resource that doesn't exist.

vincent99:
Proxying issue with more than one path segment is fixed in rancher/cattle#901.

There are some other issues to be looked at on your end:
  - Nothing calls `getImages()` so the image dropdown is empty
  - Your API returns `200 OK` with `x-error-*` headers on failure and this isn't handled (e.g. putting in the wrong credentials) because it's not a failure as far as the ajax request is concerned

andrew-ayers:
@vincent99 Just added a new commit that should correct the issues you noted; let me know if there are further changes needed. Thank you!

vincent99:
LGTM @andrew-ayers, can you squash into 1 commit (or 2 if you want to keep both you and Justin on it) and I'll merge.

ibuildthecloud:
@vincent99 @andrew-ayers Is there a list of PRs in other projects that need to be merged first?

vincent99:
@ibuildthecloud The UI won't show up until there's a schema so merging it is harmless; but [Cattle 886](https://github.com/rancher/cattle/pull/886), [Machine 8](https://github.com/rancher/machine/pull/8) (and should there be another one to add schemas for the new machine driver?)

ibuildthecloud:
@vincent99 There should be one more @wlan0 or I can do that one.

vincent99:
Merged via https://github.com/rancher/ui/pull/268

PR #272, opened 2015-10-08
Update package.json

disaster37:
Fix the issue #2263 (ember-cli-moment is deprectated to ember-moment).
With this change, npm start work fine.


vincent99:
This breaks all the places that actually use `moment`...  I imagine because it's not available as a global variable anymore.

PR #282, opened 2015-10-20
Fix Stack Collapser

ph-One:
Add missing "#" in href. Required for a functioning collapser.


vincent99:
Hmm, the `pound=true` is supposed to add that but the option doesn't make it to the helper..  But with the current Ember just adding a pound as you have is possible so we can just remove that option anyway.  Thanks

ph-One:
I thought that was strange behavior as well. Figured, if nothing else, this might draw light to a bigger issue that needs fixed. :beer:

PR #319, opened 2015-11-11
better catalogIcon display

demarant:
minor css changes. I added max-width so logos like Kibana looks better. Moreover certan svg ratio went out of the box, e.g. the logo for Plone I am adding in https://github.com/eea/rancher-catalog/tree/master/templates/plone


vincent99:
Hi @demarant, this is getting redesigned in the reskin branch (which will be master probably on Friday) and they are fixed width round-rects now:

image

demarant:
@vincent99: thanks! Indeed it looked weird with inline CSS their. I am glad see these improvements.

PR #351, opened 2015-11-19
and a bountiful pixel hunt it was

lvuch:
I went over Denise's list as well and checked off the majority of it. The dropdown position will be a little off in this.


vincent99:
On the single stack view the containers label is wrapping and the service/container count number is positioned weird..

image

PR #407, opened 2015-12-19
footer and help page changes

lvuch:


vincent99:
@lvuch This has conflicts because your fork is a few commits out of date.. @wes or I can show you how to resolve it and squash into one commit Monday.

Also the absolute footer is no good on pages that are smaller than scrollHeight, it turns on the scrollbar for no reason.

PR #424, opened 2016-01-05
cli move

lvuch:


vincent99:
@lvuch the `import Util from...` line has to move from footer to header to fix the test failure.

PR #467, opened 2016-02-04
Header add button

lvuch:
add header button


vincent99:
@lvuch I think we're holding off on the actual header add for now (and it has conflicts that you can't fix anymore because it's from your old fork), but there are some other unrelated changes to the loading screen in here.. if you want them let's put those into a separate PR

PR #481, opened 2016-02-17
fix 3600

lvuch:


vincent99:
@lvuch: `overflow: hidden` is no bueno.. When you have a short page it makes long drop downs in the header cut off.  The right fix for this is `clearfix` on the containing element of the template boxes.

ibuildthecloud:
We don't need to fix 3600, the 2600 was just fine

vincent99:
But the 5200 was cooler...

lvuch:
@vincent99 you're are right, I switched it

PR #491, opened 2016-02-26
Mobilenav

lvuch:
creating a nav for smaller screens and devices


lvuch:
@vincent99 this is ready for review you won't believe what happens next

vincent99:
@lvuch:
- Needs rebase after merging your last thing
- The Environment and [k8s] Namespace dropdowns are missing entirely when the small nav becomes active.  They need to be somewhere (maybe still in the header instead of in the hamburger, it's important context)
- k8s menu doesn't display right:
  
image

PR #493, opened 2016-02-29
Mobiletables

lvuch:


vincent99:
@lvuch: The state being an entire colored row seems a bit much?

image

PR #555, opened 2016-03-21
resolve 3088

lvuch:


vincent99:
@lvuch:
  - rancher/rancher#3660: looks fine
  - Not sure how this addresses rancher/rancher#3706 or rancher/rancher#3088
  - Update dependencies, this is trying to update to an old version of term.js

lvuch:
@vincent99 resolve #3660 and #3088 I mistyped in my original message. There may be more to 3088 but I need to get an environment with storage pools.

PR #572, opened 2016-03-29
Replace kubernetes with swarm in compose waiting template

nextrevision:
Currently seeing Kubernetes show up in the swarm section when no hosts have been added. If this isn't a typo, then you can go ahead and close this.

screen shot 2016-03-29 at 6 47 37 pm


vincent99:
Thankd

PR #585, opened 2016-04-08
Notice user that they can add container after creating stack

fyoudine-digitaslbi:


vincent99:
- Stacks have Services, not Containers.
- Saying it twice seems redundant.  Can you make it a single note that says something more like "You can always define more services later"?

fyoudine-digitaslbi:
review the UX

PR #590, opened 2016-04-11
Bugfixes

lvuch:


vincent99:
@lvuch this seems to contain a bunch of commits that aren't yours again...

PR #601, opened 2016-04-18
Feature/additional route

fyoudine-digitaslbi:
I had to hack ember's RouterDSL to reach each level of routing and keep rancher routing on top of component.


vincent99:
Merged via https://github.com/rancher/ui/pull/602

PR #610, opened 2016-04-22
Add possibilities for add-on to add entries in page-header navigation

fyoudine-digitaslbi:
I tried to implements sub-navigation with no success. However I do not have usage of this, toplevel is suffisant for my usage.


vincent99:
@fyoudine-digitaslbi we're going to be changing the design of the header a bit (back to one row with sub-menu dropdowns) so while we're doing that I'll rewrite it to render based on an input object instead of being all in templates, and add methods for you to add/remove entries.

PR #613, opened 2016-04-22
Styleguide

lvuch:


app/router.js
```diff
@@ -7,6 +7,7 @@ const Router = Ember.Router.extend({
 });
 
 Router.map(function() {
+  this.route('styleguide');
```
vincent99: This needs to be `'guide'` since the folder was renamed

PR #650, opened 2016-05-11
edit proj, vmcons, volrow

lvuch:


app/components/view-edit-project/template.hbs
```diff
@@ -2,13 +2,13 @@
   {{#if showEdit}}
     <h1>
       {{#if editing}}
-        Edit Environment: {{originalProject.displayName}}
```
vincent99: @lvuch All the {{t}}'s in this file look like they're missing quotes around the key

PR #669, opened 2016-05-17
0516issues

lvuch:


vincent99:
@lvuch Need to update your copy with master, this keeps trying to replace "{minVersion} or higher" with the old hardocded "1.9.1+"

PR #692, opened 2016-05-26
Create zh-cn.yaml

cuiyz:
Created the draft of zh-cn.yaml for Simplified Chinese translation.

### DIFF
--- a/translations/zh-cn.yaml
+++ b/translations/zh-cn.yaml
@@ -1,0 +1,2515 @@
+languageName: "Simplified Chinese (CN)"
+
+##############################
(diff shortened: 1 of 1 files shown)


vincent99:
I would suggest starting with a blank file instead and copy/paste/translating one chunk from the English file at a time.  The UI falls-back to English automatically if a key is missing, and the build-process has built-in reporting of missing translation keys that you won't get if you start from the complete English file.

vincent99:
@pengjiang80 has a first pass done, so please work together on anything incomplete/incorrect:

https://github.com/rancher/ui/pull/705

PR #711, opened 2016-06-06
fix 3190

lvuch:


vincent99:
This should only show the other target service's stack name if it's in a different stack than the service you're looking at..

{{#if (eq link.service.environmentId model.environmentID)}}
  // Show just the service name
{{else}}
  // Show stackname/servicename
{{/if}}

vincent99:
Using `generic.as`, and assuming the sentence goes "targetStack/targetService as linkName" is not great either.. these should be 4 separate translations (target is same/different, and linkName is same/different) with variables in them, something like

```yaml
serviceLink:
    withAlias:
        sameStack: "{service} as {alias}"
        differentStack: "{stack}/{service} as {alias}"
    noAlias:
        sameStack: "{service}"
        differentStack: "{stack}/{service}"
```

And 

```hbs
{{#if (eq link.service.name link.name)}}
  {{t (if (eq link.service.environmentId model.environmentID) 'serviceLink.noAlias.sameStack' 'serviceLink.noAlias.differentStack') 
    stack=link.service.displayEnvironment
    service=link.service.displayName
    alias=link.name
  }}
{{else}}
  {{t (if (eq link.service.environmentId model.environmentID) 'serviceLink.withAlias.sameStack' 'serviceLink.withAlias.differentStack') 
    stack=link.service.displayEnvironment
    service=link.service.displayName
    alias=link.name
  }}
{{/if}}

PR #721, opened 2016-06-09
4297

lvuch:


app/components/action-menu/template.hbs
```diff
@@ -2,13 +2,15 @@
 {{#if showPrimary}}
   {{#each model.primaryActions as |act index|}}
     <button class="btn btn-default btn-{{size}}" data-primary="{{index}}" aria-label="{{t act.label}}">
+      <span class="visible-xs-inline visible-sm-inline">Actions</span>
```
vincent99: These need to come from translation... `{{t 'generic.actions'}}`

PR #722, opened 2016-06-10
0610issues

lvuch:


lvuch:
@vincent99 having a bit of trouble testing storage pools but the two fixes for 4415 and 4410 should do the trick, let me know otherwise

PR #728, opened 2016-06-13
catalog and machine page adjustments

lvuch:


vincent99:
@lvuch height of the machine driver boxes?

PR #731, opened 2016-06-14
add traditional chinese language file

alfredhuang211:


vincent99:
@alfredhuang211 we will be happy to take additional languages one they are more or less completly translated, but merging this now would make a language show up that appears broken because it shows nothing but fallbacks to English.

PR #750, opened 2016-06-19
Fix word Registos --> Registros (pt-br)

frekele:


vincent99:
Thanks; if you are interested there are also a bunch of new translation keys that are missing for this language

PR #751, opened 2016-06-21
footer position fixed

liuzheng712:
the footer's position is fixed will be nice


vincent99:
This is not something we think is used enough to need to always be taking up screen real estate...

PR #754, opened 2016-06-25
v1.1-dev5 Chinese Localization Update

pengjiang80:
Update based on v1.1-dev5 changes


vincent99:
@pengjiang80 Can you squash this into one commit so the diff will show what is being changed correctly?

PR #763, opened 2016-06-29
power select rancher/rancher#4409

lvuch:
matchTriggerWidth=false may theoretically help but the issues was the width:auto but pulling this thread may unravel someone else's sweater


vincent99:
@lvuch this looks like contains part of the breakpoint PR... rebase & try again?

PR #780, opened 2016-07-14
rancher/rancher##5409 drop down fixes

lvuch:


lvuch:
fixes rancher/rancher#5430

PR #785, opened 2016-07-16
Add cookie to set default timeout

vincent99:


vincent99:
`l('service:cookies').set('timeout','60000')` in console (or some other number, in milliseconds)

PR #787, opened 2016-07-18
0718issues

lvuch:


app/styles/layout/_md-screen.scss
```diff
@@ -11,6 +11,12 @@
   .col-md-full-height {
     height: 100%;
   }
+
+  tr {
+    &.smnoborder{
```
vincent99: This is `smnoborder` here and `sm-noborder` above where it's used..

PR #788, opened 2016-07-20
Translation Japanese(ja-JP)

go:
Hi, I created ja-jp.yaml file to support Japanese(ja-JP) as new i18n support.


vincent99:
@go We're looking at switching to Crowdin for translations if you want to try it out and let us know if it's easier to work with:  http://translate.rancher.com/project/rancher-ui/invite

I imported everything from this PR.

PR #792, opened 2016-07-21
WIP french translation

xinity:
don't merge for now !

this is a WIP for french translation , any help appreciated !!

will try to finish it as soon as i can

### DIFF
--- a/translations/fr-FR.yaml
+++ b/translations/fr-FR.yaml
@@ -1,0 +1,2863 @@
+languageName: "French"
+
+##############################
+# Really generic things used in multiple places (use sparingly)
+##############################
+generic:
+  actions: Actions
+  activate: Activé
+  as: As
+  closeModal: Fermer
+  cancel: Annuler
+  created: Crée le
+  default: Defaut
+  description: Description
+  disabled: Desactivé
+  domainName: Nom de Domaine
+  emptyString: ""
+  enabled: Activé
+  expires: Expire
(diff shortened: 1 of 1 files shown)


vincent99:
@xinity We're looking at switching to Crowdin for translations if you want to try it out and let us know if it's easier to work with:  http://translate.rancher.com/project/rancher-ui/invite

I imported everything from this PR (p.s. there's a bunch of syntax errors around quoting strings and words with apostrophes in them)

xinity:
Thanks @vincent99 I've subscribed to the translate project on crowdin (using my primary email).

Apostrophe are prohibited then am I right? There is many many in french, let me know if I should not use them anymore. 

I'll try and update as fast as I can 

Thanks again, 
Rachid

vincent99:
You can have quotes, they just have to be quoted or escaped correctly to be a valid YAML file.. On the upside with something like CrowdIn you don't need to deal with the raw YAML, it will generate the file for us.

```yaml
key: thing
key: 'thing' # ok
key: 'thi'ng' # syntax error, unbalanced '
key: "thi'ng" # ok
key: "with "a quoted" string" # syntax error, unbalanced "
key: 'with "a double-quoted" string' # ok
key: |
  a multi-line string
  with unbalanced ' and "
# ^-- ok
```

vincent99:
Merged via #837, please use http://translate.rancher.com for further updates

PR #793, opened 2016-07-21
german translation for rancher ui

marsmensch:
Please find attached the complete german localization for the Rancher ui. Glad i could help, thank you for your great work on Rancher and other projects!

I want to add that i feel like the translation is way to hard at the moment (error prone, lots of duplicate strings, formatting issues) and there should be a way to make it easier for new users. 

Personally, i think something like transifex (used by botcoin for example) could boost translations significantly.


janeczku:
Hi @marsmensch, thanks a lot for contributing the German localization. I definitely see the value of using a collaborative translation platform, both with respect to ease of the process as well as the quality of the localization. I have actually just imported your translation to transifex. So maybe instead of me commenting a 3000 line file with my spell corrections and you having to edit your file manually, i could send you an invite to the transifex project and we can wrap up things there. Let me know if you are interested.

marsmensch:
Hi @janeczku, thank you for the fast feedback. Feel free to send me an invite for the transifex project and i will happily help out!

vincent99:
@marsmensch @janeczku We're looking at switching to Crowdin for translations if you want to try it out and let us know if it's easier to work with:  http://translate.rancher.com/project/rancher-ui/invite

I imported everything from this PR.

marsmensch:
Hi @vincent99, definitely easier to work with. I still see lots of duplicate strings though. For example, i would expect no further questions for "Image", "Ports" or "Services" if i translate and approve it one. The opposite is the case.

vincent99:
It doesn't change the number of tokens there are to translate, that would be in the code.  Except for a limited set of words that are very generalized each component has it's own translations.  

I know it's repetitive and there are perhaps a few more common words that could go into `generic.`; but generalizing everything and using the same token everywhere doesn't really work..  Translations depend on the context.  Upper/lowercase, 1st/3rd person, tense, synonyms, homonyms, etc.  Maybe you're trying to refer to a different concept entirely (say "[technical support] services" has a different translation than "[docker-compose] services") and it seems like a good idea to change that token... but now you've changed an unknown amount of random other places throughout the app.

marsmensch:
Hey, thanks for the detailed feedback. In general you are certainly right, but i am talking about single word occurences here. There are lots of those in the existing template, but also complete sentences that clearly have duplicates in another section. Additionally, the usage of singlular/plural forms is often inconsistent.

asdqweaa:
error:
jquery.js:9175 GET https://localhost:8000/translations/en-us.json?1.2.6 404 (Not Found)

os: win7
1: npm install bower 
2:npm install 
3: bower install --allow-root
4:npm start is ok!
whay 404?   executing ember-cli not create /dist /translations folder?
or translations server is not run?

vincent99:
Merged via #837, please use http://translate.rancher.com for further updates

PR #809, opened 2016-08-05
rancher/rancher#5643

lvuch:


app/components/form-healthcheck/template.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="row {{if editing 'form-group' ''}}">
+<div class="row form-group{{if editing 'form-group' ''}}">
```
vincent99: If it should always have form-group, remove the `{{if ...}}`

app/components/form-networking/template.hbs
```diff
@@ -54,7 +54,7 @@
 
       <div class="col-sm-12 col-md-8">
         {{#input-or-display editable=editing value=service.retainIp}}
-          <label class="form-control-static">{{input type="checkbox" checked=service.retainIp}} {{t 'formNetwork.retainIp.reuse'}}</label>
+          <p class="form-control-static">{{input type="checkbox" checked=service.retainIp}} {{t 'formNetwork.retainIp.reuse'}}</p>
```
vincent99: Nope... `<label>` is what makes the text clickable instead of just the checkbox

PR #828, opened 2016-08-24
Drone 0.5

nrvale0:
@vincent99 @cloudnautique @will-chan

**Heads up  - considerable changes in this PR!**

Drone 0.5 has a lot of new stuff and silently deprecates a lot of old stuff. The volume of this PR reflects those differences. **Note that .drone.yml syntax for Drone 0.5 is incompatible with Drone < 0.5. Also of note, the failures are often silent or give incorrect indication as to the source of failure.***

**Goal of this PR is to bring rancher/ui into the Drone 0.5 fold.**

# Required build parameters from Drone 0.5 Secrets

The following Drone Secrets are required to be set at the repo level via the ```drone secret add <name> <value>``` command. Note that one could argue that some of these things don't warrant "secret" treatment but Drone 0.5 doesn't offer a good way to parameterize / do inputs to a .drone.yml so I've over-load the Secrets subsystem for purposes of data injection. That said, I think there is a benefit to obscuring things like the name of the GCE project hosting the Cloud Storage bucket from non-insiders.

```pre
SLACK_TOKEN 
Images: plugins/slack-blame
Events: push, tag, deployment

GS_SERVICE_ACCOUNT_KEY 
Images: rancher/dind:v0.3.0
Events: push, tag, deployment

GS_SERVICE_ACCOUNT_ID 
Images: rancher/dind:v0.3.0
Events: push, tag, deployment

GS_PROJECT_NAME 
Images: rancher/dind:v0.3.0
Events: push, tag, deployment

GS_URL 
Images: rancher/dind:v0.3.0
Events: push, tag, deployment
```

Optionally, one can set a Drone Secret to throw portions of the build process into a debug mode which will, among other things, be verbose about presence of environment variables and shell variables. Be warned this will expose sensitive credential information in the Drone build output:

```pre
BUILD_DEBUG 
Images: rancher/dind:v0.3.0
Events: push, tag, deployment
```

# Significant Changes

## .drone.yml

Almost entirely re-written from previous versions. Of note:

* Drone 0.5 build asset caching requires both a drone-sftp(name?) plugin and a co-located SFTP server for hosting of assets. We don't have that infrastructure do we don't get the benefit of asset caching across builds. Also, I think push/pull of those cached assets to external SFTP is a dumb idea. Drone says they will be adding addtl caching plugins, using different transports, later but for now...no cachine.

* I've fixed Slack notifications. The target Slack channel and required token must be fed into the build as Drone Secrets.

### auth for GCS bucket
The old Drone config leveraged a GCE Service Account mapped to the VM hosting the old Drone 0.4 Agent. This created what was effectively an rhost-like authentication scheme for pushing ui releases to the GCS bucket. Meaning GCE recognized a particular GCE VM as "special" and allowed it to manipulate the contents of the bucket hosting rancher/ui release without any addtl authentication.

Because the Drone infra now lives in Packet, GCE host-based Service Accounts was no longer an option. 

build-static has been lightly modified to not only do the pushes or rancher/ui releases but first do a key/token-based auth to the GCE project at every run. Note that the required Drone Secrets are called out in the Requirements section of this PR.

The additional auth step required some rejiggering of the ./scripts directory called out below.

#### utils.sh

* Function runCmd was moved to utils.sh so it can be sourced from other scripts.
* Function gcs_asset_upload was added to utils.sh and build-static 

#### Docker Run of google/cloud-sdk has new volumes and a new CMD

Docker's handling of shell commands passed at ```docker run``` cmdline is broken by design. The Docker-based sync via Docker run which was used in the old build-static script has been adjusted in the following ways:

* A temporary file with the GCE project auth key is now bind-mounted at /tmp/keyfile.json
* the ./scripts directory is now bind-mounted to /opt/rancher/scripts so the container has access to the utils.sh and gcs_asset_upload script.
* The ```docker run``` commands have been condensed into a single command which uses /scripts/gcs_asset_upload script as a wrapper around the function of the same name (in utils.sh).


vincent99:
Can you preserve the normal `./scripts/build-static [-l] -u` behavior of using local gsutil credentials instead of requiring service accounts and credentials?  (UPLOAD=1 for human vs UPLOAD=2 for Drone)

vincent99:
And this could be a separate project, but it would be nice if it automatically did a real build in response to a tag being pushed (so I just tag and push tag instead of `./scripts/build-static -t -u` to publish a release)

nrvale0:
@vincent99 Re local credentials, that should be easy...just a matter of adding a flag to skip the auth...however that will mean that the bucket will have to be configured to allow a non-Service Account to push.

nrvale0:
@vincent99 Re the tag build, that's user-configurable on the Drone side. I just flipped  "Tag Hooks" on in the settings for the repo in Drone. All good now.

nrvale0:
@vincent99 re local credentials:

* https://github.com/nrvale0/ui/blob/drone_05/scripts/build-static#L203
* https://github.com/nrvale0/ui/blob/drone_05/scripts/utils.sh#L45

vincent99:
Those changes may make the variables not be used anymore, but they're still required..

```
vincent:ui vincent$ ./scripts/build-static
./scripts/build-static: line 6: GS_SERVICE_ACCOUNT_ID: unbound variable
./scripts/build-static: line 6: GS_SERVICE_ACCOUNT_KEY: unbound variable
```

vincent99:
OS X `mktemp` doesn't have `-p`..

```
mktemp: illegal option -- p
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

vincent99:
And some more stuff after that.. checkout the code on a mac, run `./scripts/update-dependencies`, and iterate on `./scripts/build-static -l -u -f -c "your.google.bucket/ui" -v test` until it works

nrvale0:
@vincent99 Scenarios tested and passed:

* manually given the above command line
* non-master branch in CI
* master branch in CI.

I think we are good to go assuming the CI run in Drone 0.5 passes for the current build.

PR #881, opened 2016-10-30
Some (initial) corrections on fr-fr

ThomasSchweizer:


vincent99:
@ThomasSchweizer These should go into http://translate.rancher.com/project/rancher-ui/fr please, where you can collaborate with other users of the language.  The translation files in this repo (other than the `en-us` one) are just periodically updated from CrowdIn.

ThomasSchweizer:
@vincent99 
Didn't know about this, thanks, will do :)

PR #890, opened 2016-11-14
Update ja-jp.yaml

minepicco:
I ve seen it is already translated mostly, so I just have committed only a word. "read-only"


vincent99:
@minepicco Can you please put this into http://translate.rancher.com/ , the translations files in this repo are periodically synched from there.

PR #905, opened 2016-11-21
Adds RTL support and persian language

mirmahdi90:


vincent99:
@mirmahdi90 2300 lines of flipping the left/rights on the compiled app.scss doesn't really seem sustainable.  Was it automatically or manually generated?  Automatic but manually tweaked?  Something that can be part of the build process would be much more desirable than trying to keep this in sync.

PR #925, opened 2016-11-29
Adds RTL support and persian language.

mirmahdi90:


vincent99:
@mirmahdi90 This looks much cleaner and more maintainable, thanks..  A few things to address:

  - This is unfortunately too late to get into 1.2.0, but should be doable for the next release after that.
  - We build and package all the languages, so `ember-cli-rtlcss` should be loaded all the time in package.json instead of having to manually install it.
  - After this initial import, maintenance of the language file will move to https://translate.rancher.com and be periodically synched.
  - Does the `languageName` displayed in the language dropdown need to be forced RTL regardless of the current language's direction?
    - i.e. does it display correctly if you're looking at English?
  - Instead of hardcoding `fa-ir` in various places please add and call a single function to return the direction for a given language
    - e.g. `function isRtl(lang) { return ['fa-ir'].includes(lang); }` in `user-language`.
    - This way when other RTL languages come along just that one place needs to be updated to support them.
  - `writeStyleNode()` in `user-theme` already manages the URL of the `#theme` link, so manipulating it in `user-language` `switchThemeTo()` is redundant...
    - ...and doesn't actually work correctly; the RTL will be lost if you change themes or the auto theme flips.
    - Please make `writeStyleNode()` be the one place that updates the `#theme` and `#vendor` URLs based on the current theme and language.
    - You can get the current language there with ``this.get(`session.${C.SESSION.LANGUAGE}`)``
    - And you can then call `writeStyleNode()` on the theme service inside the language service after the point where the locale is changed.
    - ([here](https://github.com/rancher/ui/blob/master/app/services/user-language.js#L64) and [here](https://github.com/rancher/ui/blob/master/app/services/user-language.js#L81)).
    - These are also the point where the displayed language is actually changed, updating the link in `setLanguage()` is after the new language is already showing so it will probably flicker and flip.

vincent99:
(p.s. you can update a PR by adding more commits, you don't need to close & open a new one)

mirmahdi90:
Thanks for your feedback, I've made the changes you mentioned.
calling `writeStyleNode()` after `setLocale()` didn't work properly, so I had to add it to `setLanguage()` as well.

vincent99:
Looks good, thanks @mirmahdi90.

PR #975, opened 2017-01-12
calculate cpu stats based on new method

StrongMonkey:
@vincent99 Use new method to calculate cpu usage percentage. Since the host stats is not from `docker stats`, we should use this method to calculate percentage. It improves the accuracy of cpu stats at some points, but we do have some backend issues to be fixed to make it more precise. Don't merge it now.


StrongMonkey:
https://github.com/rancher/rancher/issues/6895

PR #1013, opened 2017-02-06
Supported Docker version instead of latest

superseb:
As supported versions of Docker are documented, the custom host should refer to it instead of stating to install latest version.

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -2430,7 +2430,7 @@ machine:
       label: Region
 
   driverCustom:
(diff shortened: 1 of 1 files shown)


vincent99:
@superseb Windows is experimental so there is no real "supported" version.  But list is not very relevant for Windows where nothing older than ~1.12 has ever existed, and there would need to be a section in the linked docs telling you how to get the desired version of Docker in the first place.

superseb:
True, adjusted to change Linux only

PR #1067, opened 2017-03-10
implement fullscreen mode for container shell

loganhz:
hi @vincent99 @deniseschannon ,

this is a PR for container shell fullscreen mode.
i have added a fullscreen button in modal shell.
and make the fullscreen mode as default for the new window shell.

you can have a try via [link removed, don't publish unauthed installations to the internet... -- Vincent]

thanks

### DIFF
--- a/app/components/container-shell/component.js
+++ b/app/components/container-shell/component.js
@@ -1,12 +1,15 @@
 import Ember from 'ember';
 import { alternateLabel } from 'ui/utils/platform';
+import ThrottledResize from 'ui/mixins/throttled-resize';
 import Terminal from 'npm:xterm';
+import 'npm:xterm/addons/fit';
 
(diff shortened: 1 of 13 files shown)


vincent99:
This is cool, but the reason we don't already have a resizable terminal today is because it doesn't actually work.  The docker daemon has to be told to resize the TTY to match the client's size and we do not have a way to do that exposed in the API.

If you type more than 80 characters or run `top` or similar you'll see you still have an 80x24 console with a lot of black around it now.

loganhz:
yes. I found it doesn't help for command like ls.
but it do help for some case, like docker ps in rancher-agent container.
or cat a log file

loganhz:
yes, you are right. typing more than 80 characters is a big problem.  
Please decline this PR

vincent99:
Keep the branch around and we can probably eventually use it, I'll eventually get someone to put the backend support needed in.

PR #1076, opened 2017-03-16
make container shell larger

loganhz:
make shell larger and resizable.
xterm has been upgraded for bug fix and inserting international (including Chinese, Japanese, Korean) characters using IME input with keyboard.


loganhz:
a new ember-cli version 2.12.0 has been released today. 
it breaks our build i think.
it works well with ember-cli 2.11.0

PR #1079, opened 2017-03-17
upgrade xterm and make container shell larger

loganhz:
make shell larger and resizable.
xterm has been upgraded for bug fix and inserting international (including Chinese, Japanese, Korean) characters using IME input with keyboard.


vincent99:
LGTM, just need the agent merged first..

PR #1151, opened 2017-04-26
Fix typo in german language-File

justb81:


vincent99:
Please change translations in https://translate.rancher.com .  The copies in here are just periodically synced with CrowdIn for builds.

PR #1153, opened 2017-04-26
adding 17.03 and 17.04 to dropdown for install docker URL

deniseschannon:


deniseschannon:
https://github.com/rancher/rancher/issues/8138

PR #1163, opened 2017-05-03
#8646 skip volume name validation when volume driver is set in 1.6 branch

vberetti:


vincent99:
This is really too late to put in and test for 1.6.0, but we can put it in the first patch release.  I don't particularly *like* this solution, but don't have any better ideas either.

The general design is unfortunate, the validation being up to the driver creates a situation where we either need code that knows the validation rules for every possible driver individually, or just let anything through.  Which causes a long expensive failure for the user when the container won't start.

PR #1179, opened 2017-05-13
Chinese UI Translation Update

pengjiang80:
Update since v1.2


vincent99:
Can you upload this to translate.rancher.com too (for 2.0)?

pengjiang80:
Done. There are some difference and I translated that one separately.

PR #1202, opened 2017-06-01
Secrets are available in compose/cli

superseb:


vincent99:
The compose files generated by cattle are still wrong.

PR #1213, opened 2017-06-06
Fixed some french translations

gitomato:
I'm a native french speaker. I tried to use the best translations given the context. This is a long file and I can't finish it now. I will make other fixes when I see them in the UI.


vincent99:
Hi @gitomato, translations go into https://translate.rancher.com and are periodically synched into the repo.  You should be able to import this file to avoid retyping them all.

gitomato:
Thanks, I just registered.

PR #1237, opened 2017-06-24
add transplation for ecl machine catalog

mittz:
Add translation parameter for ECL machine driver catalog.

[template.hbs](https://github.com/mittz/ui-driver-ecl/blob/v1.1.0/component/template.hbs) will be applied to machine driver in community catalog after this pr is merged.

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -2725,6 +2725,21 @@ machine:
       label: SSH Password
       placeholder: Set Instance SSH Password, Blank for Auto set
 
+  driverEcl:
+    accountAccess:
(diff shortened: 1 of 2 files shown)


vincent99:
It doesn't really make any sense to put translations into the UI repo for drivers that are external.  

These should be part of the UI add-on that is configured for the driver, which you should be able [to do](https://github.com/ember-intl/ember-intl/blob/master/docs/asynchronously-loading-translations.md) with something like 

```
intl: Ember.inject.service(),

didInitAttrs() {
  this._super(...arguments);
  this.get('intl').addTranslations('ja-jp', {'your': 'translations'});
}
```

PR #1238, opened 2017-06-26
Add support display number of sidekicks for German,French,Japanese

go:


vincent99:
@go These need to [also] go into https://translate.rancher.com or they're going to be lost next time I sync them (which is only for master/2.0 at this point)

go:
@vincent99 okay, I'll try to sync my resource updates between 1.6-dev and master at crowdin.

PR #1261, opened 2017-07-18
Lv 0717

lvuch:


app/components/container/form-scale/template.hbs
```diff
@@ -11,11 +11,7 @@
   <div class="pt-10">
     {{#if showContainer}}
       <div class="radio">
-        <label>
```
vincent99: Can you revert this file, this part changes the `container` option to `global` and is no bueno...

PR #1299, opened 2017-08-15
Sort catalogs by name

vincent99:


cjellick:
addresses https://github.com/rancher/rancher/issues/9446

PR #1352, opened 2017-09-12
[2.0]Support copy host ip to the clipboard in box and dot views

loganhz:
related issue [[Feature Request] IP address copy on click](https://github.com/rancher/rancher/issues/7515)


loganhz:
@vincent99 do we need a PR for `1.6-dev` as well?

PR #1356, opened 2017-09-14
Show links in stack description

loganhz:
[related issue 5282](https://github.com/rancher/rancher/issues/5282)

image

### DIFF
--- a/app/container/template.hbs
+++ b/app/container/template.hbs
@@ -17,7 +17,7 @@
 </div>
 
 {{#if model.container.description}}
-  {{banner-message color='bg-secondary mb-0 mt-10' message=model.container.description}}
+  {{banner-message color='bg-secondary mb-0 mt-10' message=(linkify model.container.description)}}
 {{/if}}
 
 <section>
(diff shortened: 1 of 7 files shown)


loganhz:
@vincent99 please let me know if I need to apply this to other descriptions.

vincent99:
The linking logic look good but it needs to escape everything that **isn't** a link and return a `.htmlSafe()` string so that you don't need triple-`{`.

As-is this allows arbitrary HTML injection onto the page via anything shown in a banner (try setting your description to `<b>hi</b> http://google.com`), which is very bad.

Also there are lots of other descriptions (service, container, etc detail pages) that should be linkable too.

vincent99:
You can also actually load kevva/url-regex via npm instead of copying it.  Add to pacakge.json and `import urlRegex from 'npm:url-regex';`

loganhz:
@vincent99 updated.
thanks for your comments.
It looks like I can change `banner-message` component as all descriptions are using it.

vincent99:
All descriptions may be banners, but not all banners are descriptions... I haven't tried this but expect it breaks any that are given HTML input, like anything that already has an `<a href="http://blah">` in it will get double linked.  

Please leave banner as-is and linkify the input to it when used for descriptions.  Or add a `displayDescription` which does linkify to mixins/cattle-transitioning-resource and use that for the banner input.

loganhz:
@vincent99 done.

PR #1370, opened 2017-09-21
Add graph area component

loganhz:
image

[rancher/rancher#9932](https://github.com/rancher/rancher/issues/9932)

### DIFF
--- a/app/styles/components/_graph-area.scss
+++ b/app/styles/components/_graph-area.scss
@@ -1,0 +1,82 @@
+.graph-area {
+  font-size: 12px;
+
+  .line {
+    fill: none;
+    stroke: $link-color;
+    stroke-width: 2px;
+  }
+
+  .axis {
+    fill: #9da0ae;
+  }
+
+  .axis line,
+  .axis path {
+    fill: none;
+  }
+
+  .odd {
+    fill: $mid-grey;
+    fill-opacity: .1;
+  }
+
+  .even {
+    fill: $dark-grey;
+    fill-opacity: .1;
+  }
+
+  .area {
+    fill-opacity: .7;
+  }
+
+  .hover-line {
+    stroke: #9da0ae;
+  }
+
+  .hover-label {
+    position: absolute;
+    pointer-events: none;
+    top: 0px;
+    z-index: 20;
+  }
+
+  .graph-area-tooltip {
+    border-collapse: collapse;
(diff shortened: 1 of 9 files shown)


app/styles/components/_graph-area.scss
```diff
@@ -0,0 +1,82 @@
+.graph-area {
+  font: 12px Lato;
```
vincent99: `font-size: 12px;`, don't change the font-face.

app/utils/graph-tooltip.js
```diff
@@ -0,0 +1,127 @@
...
+  const ago = Math.max(0, maxPoints - index - 1) * duration / 1000;
+  if (ago === 0) {
+    return 'Now';
+  }
+
+  if (ago >= 60) {
+    const min = Math.floor(ago / 60);
+    const sec = ago - 60 * min;
+    if (sec > 0) {
+      return `${min} min, ${sec} sec ago`;
```
vincent99: (I know it was like this before, but) these should go into translation keys, and then you can delete util.js `pluralize`

app/utils/graph-area.js
```diff
@@ -0,0 +1,240 @@
...
+  params.y.rangeRound([params.height - 2, 2]);
+  params.yAxisG.call(params.yAxis);
+}
+
+function updateLines(series, params) {
+  series.forEach((serie, i) => {
+
+    serie.areaChart.attr('d', params.area(serie.data))
+      .attr('transform', null)
+      .transition()
```
vincent99: The smooth animation looks nice but is very cpu-intensive; can you make it update only when the data changes?

app/utils/graph-area.js
```diff
@@ -0,0 +1,240 @@
...
+      .each('end', () => {
+        if (i === series.length - 1) {
+          start(params)
+        }
+      });
+
+  })
+}
+
+function drawStripes(svg, width, y, yTicks) {
```
vincent99: The memory graph (and only that one, possibly because the values are the smallest integers here?) has one more axis label than yTicks suggests, so the labels don't match up with the stripes:

image

vincent99:
This looks really good overall @loganhz.

vincent99:
JS error that pops up occasionally:

image

vincent99:
Also rebase the PR please

PR #1391, opened 2017-10-09
Fix bug with error page from Up to date button

n313893254:
Add a judgement so that click button won't let machine upgrade.

### DIFF
--- a/app/mixins/upgrade-component.js
+++ b/app/mixins/upgrade-component.js
@@ -150,7 +150,7 @@ export default Ember.Mixin.create({
   doUpgrade() {
     let status = this.get('upgradeStatus');
 
-    if ( [REQUIRED,AVAILABLE,CURRENT].indexOf(status) >= 0 )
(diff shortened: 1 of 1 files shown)


vincent99:
Checking that it's not `2` is the same as not allowing `CURRENT`, i.e. `[REQUIRED,AVAILABLE,CURRENT].indexOf(status) >= 0`.

But this is specifically intentionally allowed to provide a way to do an upgrade and edit the answers on a stack without changing the version or when there is no newer version available.

n313893254:
Thank you for your review

I found the reason for this issue is  the link of  `/v2-beta/projects/1a5/stacks/1md14?sort=name` can not be found (404).  

So I am a little confused that is the API have problem or should I make the can not be click when the button is `Up to date` ?

vincent99:
Ok, that seems like a problem but should be addressed in the `upgrade-dropdown` to just have clicking anywhere in the button show the dropdown instead of having a click that does doUpgrade.  `upgrade-component` is used by other things and needs the current behavior.

PR #1393, opened 2017-10-10
Fix sortable table issue

loganhz:
[rancher/rancher#9656](https://github.com/rancher/rancher/issues/9656)


loganhz:
hi @vincent99 . 
In chrome dev tool, I can see the requests sent out when I click the header every time. 

I'm not sure if the line I changed in this PR is the root cause. but issue is gone after the change.
thanks a lot

PR #1394, opened 2017-10-10
Fix bug with error page from Up to date button

n313893254:
I replace the `icon` with `button`, so that it look like the follow picture

image
image


n313893254:
https://github.com/rancher/rancher/issues/9786

loganhz:
@n313893254 the dropdown icon is missing and you should use `if` `else` in the template instead of just deleting the button. The button is still useful when `upgradeStatus` is neither `current` nor `available`.

yasker:
@n313893254 

1. Squash into one commit.
2. Put the issue URL at the end of commit body, not in the title.

vincent99:
This seems fine; for anything you do in 1.6 check if there needs to be a corresponding 2.0 PR (there likely is in this case).

PR #1396, opened 2017-10-11
Focus the first field for add registry and webhook page

loganhz:
[rancher/rancher#3706](https://github.com/rancher/rancher/issues/3706)

### DIFF
--- a/app/registries/new/controller.js
+++ b/app/registries/new/controller.js
@@ -10,6 +10,9 @@ export default Ember.Controller.extend(NewOrEdit, {
 
   actions: {
     selectDriver: function(name) {
+      setTimeout(() => {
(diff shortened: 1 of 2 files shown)


app/registries/new/controller.js
```diff
@@ -10,6 +10,9 @@ export default Ember.Controller.extend(NewOrEdit, {
 
   actions: {
     selectDriver: function(name) {
+      setTimeout(() => {
+        $('INPUT')[0].focus();
```
vincent99: This needs to be more specific, add a class or look only in `<main>`.  There will likely be an `<input>` in the header soon.

loganhz:
@vincent99 done
thanks

PR #1402, opened 2017-10-13
[1.6]Select the first available zone which has at least one subnet.

loganhz:
Zone a was selected  by default. But sometimes there is no subnets in zone a

https://github.com/rancher/rancher/issues/9871
https://github.com/rancher/rancher/issues/5396


tfiduccia:
Version - 2.0 & 1.6 master versions 10/16
Verified fixed

PR #1403, opened 2017-10-13
Hostname Validation Change for Aliyun Machine Driver

orangedeng:
Because Aliyun changes his rules of the instance hostname, we are going to fit his rules.
Fix the reopen issue https://github.com/rancher/rancher/issues/9778.

@yasker @kingsd041

### DIFF
--- a/app/components/machine/driver-aliyunecs/component.js
+++ b/app/components/machine/driver-aliyunecs/component.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import Driver from 'ui/mixins/driver';
+import { validateHostname } from 'ui/utils/validate-dns';
 
 let ioOptimized=[
   {
@@ -518,6 +519,7 @@ export default Ember.Component.extend(Driver, {
   systemDiskCategory: systemDiskCategory,
   regions: regions,
   intl: Ember.inject.service(),
+  settings: Ember.inject.service(),
   ioOptimizedObserves: function(){
     if (this.get('aliyunecsConfig.ioOptimized')==='none'){
       this.set('instanceType',notOptimizedinstanceType);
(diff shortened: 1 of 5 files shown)


vincent99:
Do you really need Chinese characters in hostnames?  It seems likely that's going to cause trouble with Kubernetes (where it becomes the node name).

yasker:
@vincent99 agree. @orangedeng Don't allow Chinese name. It will cause a lot of trouble.

app/components/machine/driver-aliyunecs/component.js
```diff
@@ -553,13 +553,13 @@ export default Ember.Component.extend(Driver, {
         errors.push(this.get('intl').t('machine.driverAliyunecs.errors.nameTooShort'));
       }
       //check '.','-' can not be used as the first letter and the last letter
-      if (!/^[a-zA-Z0-9]([\.-]?[a-zA-Z0-9]+[\.-]?)*[a-zA-Z0-9]$/.test(name)){
+      if (!/^[a-zA-Z]([\.]?[a-zA-Z0-9\-]+[\.]?)*[a-zA-Z0-9]$/.test(name)){
```
vincent99: Is 0-9 actually not allowed by them as the first letter?  It is valid in general.  Actually looking closer this entire check is odd because it allows dots, so the hyphen checking needs to be per-label, not the beginning or end of the whole string.

I would suggest pulling this out into an `app/utils/validate-dns.js` with `validateHostname()` and `validateLabel()` functions and redoing them to be simple string manipulation functions rather than regex, other than checking for the basic allowed characters:

An individual `label`:
  - Must consist of the characters a-z (case-insensitive), 0-9, and hyphen (`str.match(/^[a-z0-9-]+$/i)`)
  - Must be 1-63 characters (str.length)
  - Cannot start with or end with a hyphen. (`str.slice(0,1) === '-'`, `str.slice(-1) === '-'`)
  - Cannot contain two consecutive hyphens. (`str.includes('--')`)
  - (CAN start with, end with, or be entirely made up of numbers)

A `hostname`:
  - Must consist of one or more valid `labels`, separated by a single dot. (`str.split(/\./)` and loop)
  - Cannot start with a dot
  - Cannot contain two consecutive dots (i.e. an empty label)
  - CAN end with a dot (this makes it an explicitly fully qualified domain name, but is not always desirable)
  - The total length of the hostname can be at most 253 characters (255 minus one for null-termination, and one for the trailing dot if not already present)

You can add simple unit tests to verify the behavior, see e.g. `tests/unit/utils/parse-port-test.js`

yasker: @vincent99 Aliyun doesn't allow hostname starting with a number, I just checked. We can only enforce the common part of different cloud provider's policy, not sure if it's a good idea.

vincent99: Yes, I'd like a general dns label validator (which can be used with other providers and other things) and then if that passes you can add additional aliyun specific checks (like this, or their max-length being 120 chars)

translations/en-us.yaml
```diff
@@ -2806,7 +2806,7 @@ machine:
       sshPasswordLengthNotValid: The length of SSH password must between eight and thirty.
       sshPasswordInvalidCharacter: SSH password contains invalid characters.
       sshPasswordFormatError: "SSH password must contain at least three out of four kinds of following characters: uppercase letter, lowercase letters, numbers, and special characters."
-
+      nameNotValidForRancher: The name is invalid according to the Rancher hostname rule.
```
vincent99: Use `{settings.appName}` instead of hardcoding Rancher here. (add `settings: Ember.inject.service()` to the component if needed)

tests/unit/utils/validate-dns-test.js
```diff
@@ -0,0 +1,29 @@
...
+    {str:"a..A0",result:false},
+    //check for dash
+    {str:"a-A0",result:true},
+    {str:"-aA0",result:false},
+    {str:"aA0-",result:false},
+    {str:"a--A0",result:false},
+    //check for specifical character
+    {str:"aA0/",result:false},
+    {str:"a/A0",result:false},
+    {str:"/aA0",result:false},
```
vincent99: Add tests for empty string, a label is too long, and the entire hostname is too long?

PR #1426, opened 2017-10-30
Fix the select render issue in Firefox

loganhz:
If there is a matching value in options,  set the value manually to avoid the Firefox issue emberjs/ember.js#15484

In chrome and Firefox, the first option will be selected automatically when the value is not in the content array. But we need to trigger the change event, otherwise the value in the model will not be updated.

https://github.com/rancher/rancher/issues/10242
https://github.com/rancher/rancher/issues/10241

### DIFF
--- a/app/components/new-select/component.js
+++ b/app/components/new-select/component.js
@@ -67,6 +67,21 @@ export default Ember.Component.extend({
     this.on('change', this, this._change);
   },
 
+  didRender: function () {
+    const selectEl = this.$()[0];
+    const value = this.get('value');
+    for (let i = 0; i < selectEl.options.length; i++) {
+      if(selectEl.options[i].value === value) {
+        selectEl.value = value;
+        break;
+      }
+    }
+
+    if (value !== selectEl.value) {
+      this._change();
+    }
+  },
+
   willDestroyElement() {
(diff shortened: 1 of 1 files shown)


vincent99:
This doesn't seem good... why is there not an option with a matching value in the first place?  If there is why is it necessarily the `[0]`th one?  If there isn't then what happens to the previous `[0]`th value?

loganhz:
Hi @vincent99 , do you mean the right fix should be putting the selected object at the beginning of the options array before we update options?

vincent99:
Depending on the situation, a `value` that is not in the `content` array may be allowed or not allowed.  So new-select can't really make the decision.  It should support either:

Not allowed (default, since this is what browsers do): 

If `this.get('value')` is not in `content.*.${optionValuePath}`, update `value` to be the value of the first item (from `ungroupedContent` if any, then `groupedContent`)

Allowed:

Add an additional entry to ungroupedContent so that an additional `<option>` will be rendered for the current value.

loganhz:
Thanks for your very detailed explanation. 
Now, it will not set the value to the `select` element if the value is invalid. 
And it will trigger the change event when the value is invalid for making the value on model updated.

PR #1427, opened 2017-10-30
Make modal in the top of screen

loganhz:
https://github.com/rancher/rancher/issues/10237

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "forever-agent": "^0.6.1",
     "glob": "^5.0.3",
     "http-proxy": "^1.11.1",
-    "lacsso": "0.0.58",
+    "lacsso": "0.0.59",
     "liquid-fire": "^0.27.3",
     "loader.js": "^4.0.10",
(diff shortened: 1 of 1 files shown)


vincent99:
This will break scrolling of a modal that is taller than the window.  (And making the modal contents scroll is no good.  Only the `<body>` should scroll.)

What is supposed to happen is the scroll position is saved, moved up to to the top to make the modal visible, and then restored on close.  You probably just need to backport https://github.com/rancher/ui/commit/0f9cc59c5e8816083049e9c5aabdace9b7f2117b

loganhz:
Thanks @vincent99 .
Can I just copy and override the modal service from add-on?

vincent99:
Defining a service that isn't referenced isn't going to do anything.. In 1.x it's in https://github.com/rancher/lacsso/blob/master/addon/services/modal.js and pulled in via NPM.

loganhz:
Thanks Vince.
PR for lacsso is https://github.com/rancher/lacsso/pull/11

PR #1430, opened 2017-11-05
Add searchable-select component

vimniky:
https://github.com/rancher/rancher/issues/10272


app/components/outside-click/component.js
```diff
@@ -0,0 +1,53 @@
...
+    document.removeEventListener('mouseup', this.handleUp, true);
+  },
+
+  handleDown(e) {
+    if (this.isDestroyed || this.isDestroying) {
+      return;
+    }
+
+    const isExcluded = this.get('excludedClasses').some((excludedClass) => {
+      return ` ${e.target.className} `.indexOf(` ${excludedClass} `) > -1;
```
vincent99: `excludedClasses` doesn't appear to be used for anything, so it can probably just be dropped.

If you want to keep it, this is not sufficient to check if `excludedClass` is a class that `e.target` has.  `excludedClasses=['a']` will match any class that contains an `a`, e.g. `foobar`.  Use `$(e.target).hasClass(excludedClass)`.

app/components/outside-click/component.js
```diff
@@ -0,0 +1,53 @@
...
+  layout,
+  isOutside: false,
+  excludedClasses: [],
+
+  onOutsideClick() {},
+
+  init() {
+    this._super(...arguments);
+    this.handleDown = this.handleDown.bind(this);
+    this.handleUp = this.handleUp.bind(this);
```
vincent99: Save these to a separate variable (like `_boundHandleDown`) instead of overwriting the original name and leaving no direct reference to it.

app/components/searchable-select/component.js
```diff
@@ -0,0 +1,287 @@
...
+
+  // input
+  value: null,
+  prefix: null,
+  suffix: null,
+  prompt: null,
+  class: null,
+  // If need to catch the group changes, you can pass a group prop in.
+  group: null,
+  content: [],
```
vincent99: `content: null`.  The user is going to overwrite it anyway so it doesn't matter much here, but if they don't then _every_ instance of this component will share the _same_ array when written like this and cause weird interaction between instances.

If you ned to set the default for a field to an object or array, set it in `init()` so each instance gets it's own copy.

vimniky:
@vincent99 Updated, according to your suggestions.

PR #1441, opened 2017-11-13
Support region options in lb

loganhz:
https://github.com/rancher/rancher/issues/10280
https://github.com/rancher/rancher/issues/10328


app/components/form-balancer-rules/component.js
```diff
@@ -34,6 +37,9 @@ export default Ember.Component.extend({
     protos.sort();
     this.set('protocolChoices', protos);
 
+    const regions = this.get('regions').get('all');
+    this.set('hasRegion', regions.content.length > 0)
```
vincent99: `regions.get('length')`.. Store returns an ArrayProxy, you never deal with the `content` of it directly.

app/authenticated/route.js
```diff
@@ -224,6 +226,14 @@ export default Ember.Route.extend(Subscribe, PromiseToCb, {
     }
   },
 
+  loadRegions() {
+    let svc = this.get('regions');
+    return svc.getAll().then((all) => {
```
vincent99: I don't think you need a service for this, `projects` only has one because it has a bunch of other logic for picking a default and stuff.  `regions: ['userSchemas', this.cbFind('region', userStore')]`

PR #1450, opened 2017-11-27
Implemented CRUD for project

loganhz:


lib/shared/addon/components/new-edit-project/component.js
```diff
@@ -0,0 +1,114 @@
...
+  },
+
+  doSave() {
+    return this._super.apply(this, arguments).then((project) => {
+      const projectId = project.id;
+      const members = this.get('memberArray');
+      const promises = [];
+      members.forEach(member => {
+        member.projectId = projectId;
+        const promise = this.get('authzStore').rawRequest({
```
loganhz: what's the correct way to save a list of role bindings? is there any ember store api I can use?

vincent99: You're using it right here? Make the members binding records with authzStore.createRecord({type: "projectRoleTemplateBinding"})`in addMember and `save()` then here.

We may add an action like "setMembers" because I don't like having to do N+1 saves... But this will work for now.

app/authenticated/projects/edit/route.js
```diff
@@ -0,0 +1,16 @@
+import { hash } from 'rsvp';
+import { inject as service } from '@ember/service';
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+    authzStore: service('authz-store'),
+    model: function (params) {
+        return hash({
+            project: this.get('authzStore').find('project', params.project_id),
+            projectRoleTemplateBindings: this.get('authzStore').findAll('projectRoleTemplateBinding', { forceReload: true })
```
vincent99: If you update to what I just pushed and update dependencies (ember-api-store), there's a `hasMany()` in `ember-api-store/utils/denormalize` so you can go to the project model and add

```
import { hasMany } from 'ember-api-store/utils/denormalize';

... Resource.extend({
...
  roleTemplateBindings: hasMany('id', 'projectRoleTemplateBinding', 'projectId'),
...
```

and then each `project` resource will have a live array of the associated `projectRoleTemplateBinding` resources.  You will need to preload them all (in `app/authenticated/route.js`, look for `this.preload(`)

app/authenticated/projects/index/template.hbs
```diff
@@ -0,0 +1,14 @@
+<section class="header clearfix">
+  <h1 class="pull-left">{{t 'projectsPage.header'}}</h1>
+  <div class="vertical-middle"></div>
+
+  <div class="right-buttons">
+    {{#link-to "authenticated.projects.new" class="btn bg-primary btn-sm icon-btn"}}
+    <span class="darken">
```
vincent99: Indent body of link-to block

app/authenticated/projects/edit/route.js
```diff
@@ -0,0 +1,16 @@
+import { hash } from 'rsvp';
+import { inject as service } from '@ember/service';
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+    authzStore: service('authz-store'),
```
vincent99: Not sure what editor you're using but in this project indenting is 2 spaces (not 4, or a tab)..

app/authenticated/projects/new/route.js
```diff
@@ -0,0 +1,25 @@
+import { hash } from 'rsvp';
+import { inject as service } from '@ember/service';
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+    authzStore: service('authz-store'),
```
vincent99: 4 space tabs again, but in only some files?

app/authenticated/projects/new/route.js
```diff
@@ -0,0 +1,25 @@
...
+    model: function () {
+        const project = this.get('authzStore').createRecord({
+            type: `project`,
+            name: '',
+        });
+        const projectRoleTemplateBindings = [{
```
vincent99: Default the list to empty here (and call addMember after render in the component if you want an empty row).  Doing it here creates a duplicate copy of what an empty binding looks like.

app/models/project.js
```diff
@@ -85,21 +75,25 @@ var Project = Resource.extend(PolledResource, {
...
   }),
 
+  delete: function (/*arguments*/) {
+    var promise = this._super.apply(this, arguments);
+    return promise.then(() => {
+      this.set('state', 'removed');
+    }).catch((err) => {
+      this.get('growl').fromError('Error deleting', err);
+    });
+  },
```
vincent99: Something still probably needs to happen to handle the case where you delete the project that was last selected, since the subscribe connection will still be active..  Reloading was a simple way to handle it for this edge-ish case.

lib/shared/addon/components/new-edit-project/component.js
```diff
@@ -0,0 +1,148 @@
...
+
+  primaryResource: alias('model.project'),
+  memberArray: alias('model.projectRoleTemplateBindings'),
+
+  actions: {
+    cancel() {
+      this.goBack();
+    },
+    addMember(kind) {
+      this.get('memberArray').pushObject({
```
vincent99: These should be store.createRecords() to make actual models

lib/shared/addon/components/new-edit-project/component.js
```diff
@@ -0,0 +1,148 @@
...
+
+    if ((this.get('model.project.name') || '').trim().length === 0) {
+      errors.push(this.get('intl').findTranslationByKey('projectsPage.new.errors.nameReq'));
+    }
+
+    if (!this.get('editing') && this.doesNameExist()) {
+      errors.push(this.get('intl').findTranslationByKey('projectsPage.new.errors.nameInExists'));
+    }
+
+    if (this.doseMemberNameInvalid()) {
```
vincent99: And then if the members were models most of this could be taken care of with the standard api schema validation (I dunno if the rules are setup right yet, but they should/will be).

lib/shared/addon/components/new-edit-project/component.js
```diff
@@ -0,0 +1,148 @@
...
+  setMembers(project) {
+    const projectId = project.id;
+    const members = this.get('memberArray');
+    const promises = [];
+    // TODO - wait for setMembers API suppot.
+    return this.get('authzStore').findAll('projectRoleTemplateBinding', { forceReload: true })
+      .then(bindings => {
+        const currentBindings = bindings.filter(b => b.projectId === projectId);
+
+        currentBindings.forEach(currentMember => {
```
vincent99: Do the adds before the deletes (though this won't matter if we add a setMembers() action).

Doing delete and then add could result in locking yourself out by deleting the binding that's allowing you add bindings and then not being able to add the rest.  (i.e. if you had "Logan has role Admin", removed that, and added "[Group Logan is in] has role Admin")

lib/shared/addon/components/new-edit-project/component.js
loganhz: `projectRoleTemplateBinding.validationErrors()` returns an empty array when subjectName is empty. `subjectName` is nullable in schema

PR #1453, opened 2017-11-29
Add Pipeline tab to Rancher UI 1.6

biblesyme:
## Why do we need this change?

We need a access point for Pipeline UI in 1.6 like k8s dashboard.

Related issus is [here](https://github.com/rancher/rancher/issues/10332)

### DIFF
--- a/app/components/page-header/component.js
+++ b/app/components/page-header/component.js
@@ -80,7 +80,9 @@ export default Ember.Component.extend(HoverDropdown, {
       if ( typeof item.alertCondition === 'function' && item.alertCondition.call(this) === true ) {
         item.showAlert = true;
       }
-
+      if ( typeof item.alertLink === 'function') {
+        item.alertLink = item.alertLink.call(this);
+      }
       item.submenu = (item.submenu||[]).filter((subitem) => {
         if ( typeof subitem.condition === 'function' && !subitem.condition.call(this) ) {
           return false;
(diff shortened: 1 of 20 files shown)


vincent99:
What does this accomplish different from the last PR?  1.6-dev already expects an API with region support and will not work without it, adding your code branched from and old tag just creates conflicts, the region code is still here.

biblesyme:
I'm kind of confused, I was told I need to integrate Pipeline tab part into rancher/server:v1.6.12. I assume I should develop Pipeline tab based on v1.6.10( since we now use the rancher/server:v1.6.10) so that there won't be region problem when Tani test it.

Whatever, since I wanna make Pipeline tab released along with rancher/server:v1.6.12, there are three things I am not sure.

1. Which version of rancher/ui should I based on to develop the Pipeline tab

2. Which branch of rancher/ui should I request the PR to.

3. For test, Which version of rancher/server should I use?

vincent99:
1. 1.6-dev.  working off an old tag and putting in a PR to 1.6-dev just means your code is out of date (which is why there's a conflict here), it doesn't make all the newer code in the target branch disappear.  You can _develop_ against an old tag (which as I said is UI **1.6.29** for Rancher 1.6.10) and server in your own branch, but for a PR both features need to work together.

2. 1.6-dev.

3. Master or wherever the regions API support is, according to the program working on that.

yasker:
@biblesyme Need to discuss this more.

PR #1460, opened 2017-11-30
Add new tab as service access point to Rancher UI 1.6

biblesyme:
# Implement a generic mechanism to support any service to add a new header tab as its access point

## Why do we need this?
To support Pipeline or other services's access point in the UI

## How it works?
User need to add three label to the service which need access point in page header.

1. `io.rancher.service.ui_link.label` -- This is the label show in the page header. This should be a JSON map used for i18n. it's like 
  ```JSON
  {"de-de":"Deutsch","en-us":"English","fa-ir":"فارسی","fr-fr":"Française","hu-hu":"Magyar","ja-jp":"日本語","none":"None","pt-br":"Português","ru-ru":"Русский","uk-ua":"Українська","zh-hans":"简体中文"}
  ```
	
2.  `io.rancher.service.ui_link.path` -- This is the relative router in it's own app. Default will be '/'.
	
3. `io.rancher.service.ui_link.port` -- This is the private port where the UI is hosted

All of these three label will be combined as access url for magic_proxy, as `${window.location.origin}${this.get('app.magicEndpoint')}/projects/${this.get('projectId')}/${serviceName}/${serviceAppUI}`.
The label `io.rancher.service.ui_link.label`  is the key label to detect whether this service need access point or not. The other two labels is optional.

Related issus is [here](https://github.com/rancher/rancher/issues/10332)

### DIFF
--- a/app/models/service.js
+++ b/app/models/service.js
@@ -12,6 +12,7 @@ var Service = Resource.extend({
 
   instances: denormalizeIdArray('instanceIds'),
   instanceCount: Ember.computed.alias('instances.length'),
+  projectId: Ember.computed.alias(`tab-session.${C.TABSESSION.PROJECT}`),
   stack: denormalizeId('stackId'),
 
   actions: {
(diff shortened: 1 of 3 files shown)


app/models/service.js
```diff
@@ -400,6 +401,29 @@ var Service = Resource.extend({
       return Util.formatSi(this.get('launchConfig.memoryReservation'), 1024, 'iB', 'B');
     }
   }),
+
+  serviceApp: function(){
+    let labels = this.get('launchConfig.labels');
+    let serviceName = this.get('name');
+    let serviceAppUI = labels[C.LABEL.SERVICE_APP_UI] || '';
```
vincent99: This is an odd name for what is essentially the path for the service and would often be `/` so it could just be optional.

app/utils/constants.js
```diff
@@ -186,6 +186,9 @@ var C = {
     LAUNCH_CONFIG: 'io.rancher.service.launch.config',
     LAUNCH_CONFIG_PRIMARY: 'io.rancher.service.primary.launch.config',
     SIDEKICK: 'io.rancher.sidekicks',
+    SERVICE_APP_UI: 'io.rancher.service.app.ui',
```
vincent99: Group related labels with a common prefix, I.e. `io.rancher.service.ui_link.{label, port, path}`.

vincent99: You may want to support the label being a JSON map of language code to label, or else you won't be able to have an app label localized for English vs China vs other..

app/models/service.js
```diff
@@ -400,6 +401,29 @@ var Service = Resource.extend({
...
+    let serviceAppUIPort = labels[C.LABEL.SERVICE_APP_UI_PORT] || '';
+    let serviceAppName = labels[C.LABEL.SERVICE_APP_NAME] || '';
+    let url;
+    if( serviceAppName ){
+      url = `${window.location.origin}${this.get('app.magicEndpoint')}/projects/${this.get('projectId')}`;
+      if(!serviceAppUIPort){
+        url += `/${serviceName}/${serviceAppUI}`;
+      }else{
+        url += `/${serviceName}:${serviceAppUIPort}${serviceAppUI}`;
```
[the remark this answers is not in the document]
biblesyme: Maybe Slash is redundant.  Cos here the `${serviceAppUI}` has had the slash

app/utils/constants.js
biblesyme: Shall we make the `io.rancher.service.ui_link.label` optional between JSON map and string, or it's must be a JSON map?

app/models/service.js
```diff
@@ -400,6 +401,58 @@ var Service = Resource.extend({
...
+  serviceApp: function(){
+    let labels = this.get('launchConfig.labels');
+    let serviceName = this.get('name');
+    let serviceUIPath = labels[C.LABEL.SERVICE_UI_PATH] || '/';
+    let serviceUIPort = labels[C.LABEL.SERVICE_UI_PORT] || '';
+    let localizedServiceUILabel = this.get('localizedServiceUILabel');
+    let url;
+    if( localizedServiceUILabel ){
+      url = `${window.location.origin}${this.get('app.magicEndpoint')}/projects/${this.get('projectId')}`;
+      if(!serviceUIPort){
```
vincent99: I'll merge it anyway but this is a pet-peeve.. do `if ( condition ) { a } else { b }` instead of `if ( !condition ) { b } else { a }`

biblesyme: Sorry, I will be careful next time.

PR #1472, opened 2017-12-06
Filter driver `forwardPost` of webhook-service in UI

biblesyme:
Since the [PR](https://github.com/rancher/webhook-service/pull/66) to add new driver `forwardPost` is merged.

We currently just keep `forwardPost` for internal usage. So, users should not be able to see it.

Related Issue is [here](https://github.com/rancher/rancher/issues/10479)

### DIFF
--- a/app/authenticated/project/api/hooks/index/controller.js
+++ b/app/authenticated/project/api/hooks/index/controller.js
@@ -4,7 +4,10 @@ import Sortable from 'ui/mixins/sortable';
 export default Ember.Controller.extend(Sortable, {
   settings: Ember.inject.service(),
 
-  sortableContent: Ember.computed.alias('model.receivers'),
+  sortableContent: function() {
+    let receivers = this.get('model.receivers');
+    return receivers.filter(ele => ele.driver !== 'forwardPost');
+  }.property('model.receivers.@each.driver'),
   sortBy: 'name',
   sorts: {
     state:        ['stateSort','name','id'],


app/authenticated/project/api/hooks/route.js
```diff
@@ -13,6 +13,8 @@ export default Ember.Route.extend({
 
       return Ember.RSVP.hash({
         receivers: this.get('webhookStore').findAll('receiver', {forceReload: true}),
+      }).then(({receivers}) => {
+        return receivers.filter((ele) => ele.driver !== 'forwardPost');
```
vincent99: This makes it a static array instead of a live one that will update when receivers come and go while the page is up.  Filter as a computed property in the controller instead of the route.

biblesyme: @vincent99 I've updated it.

app/authenticated/project/api/hooks/index/controller.js
```diff
@@ -4,7 +4,10 @@ import Sortable from 'ui/mixins/sortable';
 export default Ember.Controller.extend(Sortable, {
   settings: Ember.inject.service(),
 
-  sortableContent: Ember.computed.alias('model.receivers'),
+  sortableContent: function() {
+    let receivers = this.get('model.receivers');
+    return receivers.filter(ele => ele.driver !== 'forwardPost');
+  }.property('model.receivers.@each'),
```
vincent99: @each needs a key name after it, just @each is invalid.  Or you meant `model.receivers.[]`, but it should be `model.receivers.@each.driver` here.

biblesyme: Sorry for that, I meant to `model.receivers.[]`, but use it wrong, I thought the property `driver` won't be changed. 
But, I think your are right, using `model.receivers.@each.driver` is more accurate, I've updated to `model.receivers.@each.driver`.

PR #1475, opened 2017-12-07
[1.6]Enhance Windows Environment Experience

orangedeng:
refer to https://github.com/rancher/rancher/issues/10442
and https://github.com/rancher/rancher/issues/10448

1. Change the add host step for windows environment. Add setting
subnet labels logic.
2. Run command in Windows environment is fetched from rancher
server.
3. Block the webhook link on windows. Because the webhook is not
fully supported in Windows environment.
4. Because the metadata and DNS services are supported, service
alias and external service can be use in Windows environment.

### DIFF
--- a/app/components/machine/driver-custom/component.js
+++ b/app/components/machine/driver-custom/component.js
@@ -1,12 +1,15 @@
 import Ember from 'ember';
 import ManageLabels from 'ui/mixins/manage-labels';
 import Util from 'ui/utils/util';
+import C from 'ui/utils/constants';
 
 export default Ember.Component.extend(ManageLabels, {
   settings      : Ember.inject.service(),
   projects      : Ember.inject.service(),
   cattleAgentIp : null,
   model         : null,
+  subnet        : null,
+  routerIp      : null,
 
   actions: {
     cancel() {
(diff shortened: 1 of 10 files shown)


orangedeng:
@yasker

app/components/machine/driver-custom/component.js
```diff
@@ -79,11 +110,21 @@ export default Ember.Component.extend(ManageLabels, {
...
-
-    return `New-Item -Path 'C:\\Program Files\\rancher' -Type Directory
-Invoke-WebRequest -UseBasicParsing 'https://github.com/rancher/agent/releases/download/v0.6.0/agent.exe' -OutFile 'C:\\Program Files\\rancher\\agent.exe'
-& 'C:\\Program Files\\rancher\\agent.exe' -register-service ${url}
-Start-Service rancher-agent`;
-  }.property('model.command'),
+    let cattleIp = this.get('cattleAgentIp');
+    let env = Util.addQueryParams('', this.get('model.labels')||{});
+    let cmd1=`New-Item -Path 'C:\\Program Files\\rancher' -Type Directory`;
+    let cmd2 = this.get('model.windows-command');
```
vincent99: Please tell me this is a typo and there's not really a field called `windows-command` in the API?  JSON fields are camelCase, hyphen is ambiguous and requires quoting in people's code

app/components/machine/driver-custom/component.js
```diff
@@ -16,12 +18,41 @@ export default Ember.Component.extend(ManageLabels, {
...
+        if(!labels){
+          labels={};
+        }
+        for(var tmp in labels){
+          out[tmp]=labels[tmp];
+        }
+        var subnet=this.get('subnet');
+        var routerIp=this.get('routerIp');
+        if (subnet) {
+          out['io.rancher.network.per_host_subnet.subnet']=subnet;
```
vincent99: Put into constants (and the other below)

app/components/machine/driver-custom/component.js
```diff
@@ -16,12 +18,41 @@ export default Ember.Component.extend(ManageLabels, {
     setLabels(labels) {
       if ( this.get('model') ) {
         var out = {};
+        var oldlabels=this.get('model.labels');
```
vincent99: `const out = Object.assign({}, this.get('model.labels'));` (and below)

app/components/machine/driver-custom/component.js
deniseschannon: @orangedeng let's change this to `windowsCommand`

PR #1509, opened 2017-12-25
Scheduling

loganhz:


lib/shared/addon/components/container/form-scheduling/component.js
```diff
@@ -27,96 +19,74 @@ export default Component.extend(ManageLabels, {
...
+        item.toggleProperty('expanded');
       });
     }
   },
 
-  updateLabels(labels) {
-    this.sendAction('setLabels', labels);
+  loadHosts() {
+    const clusterId = this.get('scope.currentCluster.id');
+    this.get('globalStore').findAll('machine', { filter: { clusterId: this.get('scope.currentCluster.id') } }).then((hosts) => {
```
vincent99: Machines are already loaded by one of the parent routes and this is creating a static (non-live) array..  Get a copy of all machines and then add a computed property that depends on it, e.g.

```javascript
_allMachines: null,
init() {
  ...
  set(this, '_allMachines', get(this, 'globalStore').all('machine'));
  ...
}

hostChoices: computed('_allMachines.@each.clusterId', function() {
  return get(this, '_allMachines').filterBy('clusterId', get(this, 'scope.currentCluster.id'));
})

File: lib/shared/addon/components/container/form-scheduling/component.js [hunk 2]
```diff
@@ -27,96 +19,74 @@ export default Component.extend(ManageLabels, {
...
       this.set('requestedHostId', null);
+      delete scheduling['nodeId'];
     }
   }.observes('isRequestedHost'),
 
-  requestedHostIdDidChange: function() {
+  requestedHostIdDidChange: function () {
     var hostId = this.get('requestedHostId');
-    this.sendAction('setRequestedHost', hostId);
+    this.set('scheduling.nodeId', hostId);
```
vincent99: `scheduling.node.nodeId` (and elsewhere, the workload.nodeId  is/will be readonly, workload.scheduling.node.nodeId is where you request one)

app/container/template.hbs
```diff
@@ -103,10 +103,9 @@
 
     <div class="mt-20">
       {{container/form-scheduling
-        initialHostId=model.container.requestedHostId
-        initialLabels=model.container.labels
+        scheduling=model.container.scheduling
+        initialHostId=model.container.scheduling.nodeId
```
vincent99: `scheduling.node.nodeId`

lib/shared/addon/components/node-selector/template.hbs
```diff
@@ -0,0 +1,81 @@
...
+                content=operatorChoices
+                localizedLabel=true
+                value=rule.operator
+                readOnly=editing
+                style="display: inline-block; width: auto;"
+              }}
+            </td>
+            <td>&nbsp;</td>
+            <td>
+              {{#if (and (not-eq rule.operator 'Exists') (not-eq rule.operator 'DoesNotExist'))}}
```
vincent99: Add computed properties for these and flip the order, they don't need to be nested conditions..

```
{{#if isJustLabel}}
  generic.na
{{else if isMultiple}}
  input with multiple placeholder
{{else}}
  input with no placeholder
{{/if}}
```

```
  isJustLabel: computed('rule.operator', function() {
    return ['Exists','DoesNotExist'].includes(get(this, 'rule.operator'));
  }),
  isMultiple: computed('rule.operator', function() {
    return ['In','NotIn'].includes(get(this, 'rule.operator'));
  }),
```

lib/shared/addon/components/scheduling-toleration/component.js
```diff
@@ -0,0 +1,95 @@
+import Component from '@ember/component';
+import { observer } from '@ember/object'
+import layout from './template';
+const operatorChoices = ['Equals', 'Exists'];
```
vincent99: Translations like in scheduling labels

translations/en-us.yaml
```diff
@@ -2260,8 +2269,34 @@ formScheduling:
...
+  scheduler:
+    label: Scheduler
+    placeholder: e.g. myScheduler
+  nodeSelector:
+    requireAll: "Require ALL of:"
+    requireAny: "Require Any of:"
+    preferred: "Prefer Any of:"
+    operator:
+      lt: '<'
+      le: '<='
```
vincent99: le: `&le;`
ge:  `&ge;`
ne: `&ne;`

lib/shared/addon/utils/constants.js
```diff
@@ -238,6 +238,49 @@ var C = {
     ]
   },
 
+  SCHED_NODE_SELECTOR_OPERATOR: [
+    {
+      value: '<',
+      label: 'lt',
```
vincent99: Just set this to a full translation key (`formScheduling.nodeSelector.operator.(label)`) key here and then use the constant instead of computing a similar map in `initOperatorChoices`

app/container/template.hbs
loganhz: the type of nodeId is `reference[node]` via (`/v3/projects/project-${id}/schemas/nodeScheduling`). 
Should I use something like `scheduling.node.nodeId.id` here?

PR #1554, opened 2018-01-14
RBAC

loganhz:


loganhz:
PR for ember-api-store

https://github.com/rancher/ember-api-store/pull/15

I have bumped the version from `2.5.3` to `2.5.6`, maybe we need to merge other ember-api-store PR first.

vincent99:
ember-api-store 2.5.6 is published now.

- Rename `rbac` input to `resource`, that's what it is...

- Reuse `list` and `create` verbs for `permission` values matching the `canX` calls in store instead of making up new less specific words (`read` and `write`).

- Can has-permission be a helper instead of a component?  Helpers are singleton and there's a decent amount of overhead to creating component every time (also for disabling, below)

- For page-header removing is good.

- For buttons ("Deploy", "Add Secret", etc), disable the button instead of removing entirely so it's more explicit there is a thing that _can_ be done but you can't do it.

- For disabled, an opposite version of the helper would be useful.  Like `rbac-allows` vs `rbac-prevents`.  Putting all above together:

`<button disabled={{rbac-prevents resource="workload" permission="create"}}>Deploy</button>`

```
{{#if (rbac-allows scope="global" resource="cluster" permission="list")}}
  {{link-to-as-attrs ...}
{{/if}}
```

loganhz:
@vincent99 how to inject service into a helper?

loganhz:
@vincent99 done

loganhz:
@vincent99  what should we do for the action buttons like `edit` and `delete` ?
https://github.com/rancher/rancher/issues/1990

vincent99:
Edit and delete already look for the presence of the `update` and `remove` links and hide themselves if they are not allowed on a particular resource.  (which is not super consistent, there used to be too many actions to show them all grayed out but we can maybe revisit that).

Can you make `rbac-prevents` import the `compute` logic from the `-allows` and return !that instead of duplicating all the logic?  Otherwise looks good

loganhz:
@vincent99 updated

PR #1555, opened 2018-01-15
Fix

loganhz:
https://github.com/rancher/rancher/issues/10843
https://github.com/rancher/rancher/issues/10847
https://github.com/rancher/rancher/issues/10845
https://github.com/rancher/rancher/issues/10851
https://github.com/rancher/rancher/issues/10836
https://github.com/rancher/rancher/issues/10855
https://github.com/rancher/rancher/issues/10853
https://github.com/rancher/rancher/issues/10850
https://github.com/rancher/rancher/issues/10835
https://github.com/rancher/rancher/issues/10179

### DIFF
--- a/app/authenticated/cluster/index/route.js
+++ b/app/authenticated/cluster/index/route.js
@@ -10,8 +10,7 @@ export default Route.extend({
       const cluster = this.get('scope').currentCluster;
 
       return {
-        description: cluster.description,
-        displayName: cluster.name,
+        cluster,
         nodes,
       };
     });
--- a/app/authenticated/cluster/index/route.js
+++ b/app/authenticated/cluster/index/route.js
@@ -6,11 +6,12 @@ export default Route.extend({
   scope: service(),
 
   model: function () {
-    return this.get('globalStore').findAll('machine', { filter: { clusterId: this.get('scope.currentCluster.id') } }).then((nodes) => {
+    return this.get('globalStore').findAll('machine').then((nodes) => {
       const cluster = this.get('scope').currentCluster;
+
       return {
         displayName: cluster.name,
-        nodes,
+        nodes: nodes.filter(n => n.clusterId === cluster.id),
       };
     });
   },
--- a/app/authenticated/cluster/index/route.js
+++ b/app/authenticated/cluster/index/route.js
@@ -11,7 +11,7 @@ export default Route.extend({
 
       return {
         displayName: cluster.name,
-        nodes: nodes.filter(n => n.clusterId === cluster.id),
+        nodes,
       };
     });
   },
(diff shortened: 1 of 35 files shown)


lib/global-admin/addon/clusters/new/cloud/controller.js
```diff
@@ -38,6 +39,22 @@ export default Controller.extend(NewOrEdit, {
     ];
   }),
 
+  validate() {
```
vincent99: Can you remove the commit for this PR?  The server-side should be doing unique checks and returning 422 on POST if it's non-unique.  Checking in the client is error prone (race conditions, or a cluster you don't have access to see exists).  Just make sure if there's an error returned from the API that it's surfaced to the user (I think that's the actual problem in rancher/rancher#10855).

(also as of a day or two ago cluster id is not actually the name anymore so the name does not really have to be unique)

app/authenticated/cluster/index/route.js
```diff
@@ -6,11 +6,12 @@ export default Route.extend({
...
 
   model: function () {
-    return this.get('globalStore').findAll('machine', { filter: { clusterId: this.get('scope.currentCluster.id') } }).then((nodes) => {
+    return this.get('globalStore').findAll('machine').then((nodes) => {
       const cluster = this.get('scope').currentCluster;
+
       return {
         displayName: cluster.name,
-        nodes,
+        nodes: nodes.filter(n => n.clusterId === cluster.id),
```
vincent99: This needs to be a live array (from `findAll()`) that is then filtered as a computed property in the controller; Any time you do a filter in the route you no longer have a live collection and it will not reflect new or removed nodes as they change while you're still on the page.

lib/shared/addon/components/form-name-description/component.js
```diff
@@ -18,6 +18,7 @@ export default Component.extend({
   nameHelpText           : '',
   nameRequired           : false,
   nameDisabled           : false,
+  horizontal             : false,
```
vincent99: Can you just tweak the add account form to use the normal (horizontal=false) view so that it's more consistent with other pages?

That way you don't need to add this `horizontal` option (which doesn't support all the description-related options).

loganhz:
@vincent99 updated

app/authenticated/cluster/index/route.js
```diff
@@ -10,6 +10,7 @@ export default Route.extend({
       const cluster = this.get('scope').currentCluster;
 
       return {
+        description: cluster.description,
```
vincent99: This already exists so you didn't cause it but this is another place where the route is returning non-live data that won't auto update if it changes... return the `cluster` itself and then show `model.cluster.description` instead.

Also if possible (I'm not sure it has all the same data?) use `machines` instead of `nodes` for the dashboard info.  We're trying to not expose 2 slightly different types for these to the public API.  The `machines` of a `cluster` are also a computed property on `cluster` so you don't need to do anything special in the route to get them.

loganhz:
@vincent99 done. Thanks for your review!
I'm using the `machine` type for dashboard.

PR #1571, opened 2018-01-18
[1.6]Webhook for Windows environment

orangedeng:
Related issue https://github.com/rancher/rancher/issues/10663
Webhook is enable for Windows.
Webhook only can be scale service or upgrade a service in Windows environment because there is no any machine driver for windows.


orangedeng:
@deniseschannon need to be reviewed.

PR #1609, opened 2018-01-29
Fixed: In grouping mode searchable-select takes two click to select a…

vimniky:
In grouping mode, it takes two clicks to select a searchable-select option.

Here's the related link:
https://stackoverflow.com/questions/9335325/blur-event-stops-click-event-from-working

### DIFF
--- a/lib/shared/addon/components/searchable-select/template.hbs
+++ b/lib/shared/addon/components/searchable-select/template.hbs
@@ -28,7 +28,7 @@
       {{#each unGroupedContent as |item idx|}}
         <div
           class="searchable-option"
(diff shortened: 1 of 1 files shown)


vincent99:
I haven't checked what this does on grouped dropdowns to see if it's different, but this is weird on regular ones.  Form components operate on mouse up, not mouse down.

vimniky:
I still can't figure out why It doesn't work on grouped dropdowns but when I change event from `click` to `mouseDown` the bug got fixed. I will look it again to see if there is a better solution.

vincent99:
Closing for now, changing the regular behavior from mousedown to up is not good...

PR #1616, opened 2018-02-02
RKE config

loganhz:


vincent99:
@loganhz I pulled all these options out, we're going to have a single k8s version that will cause Rancher to select the full set of images that will be deployed, and a separate option for what registry to get them from.  Not expose 20 different knobs for the user to configure.

loganhz:
@vincent99 got it.

Do we need the options for `clusterCidr` and `serviceClusterIpRange` of `kubeController` in UI?

vincent99:
Yes you can add those.  Also the option to allow unsupported docker versions.

PR #1629, opened 2018-02-08
exoscale: zones and templates

greut:
Hi Rancher,

A little update to the exoscale driver letting the customer choosing a data-center as well as a template.

screenshot from 2018-02-02 14-34-48

screenshot from 2018-02-02 14-35-08

screenshot from 2018-02-02 14-35-34

Cheers,


vincent99:
FYI you can do `{{if (eq step 5)}}` and similar with `ember-truth-helpers` in modern versions, instead of having lots of different `isStep[Gte]N` properties.

greut:
@vincent99 thanks for the merge and the tip!

PR #1631, opened 2018-02-10
Fix pod exec and log issue

loganhz:
https://github.com/rancher/rancher/issues/11051

### DIFF
--- a/lib/shared/addon/components/container-shell/component.js
+++ b/lib/shared/addon/components/container-shell/component.js
@@ -40,7 +40,7 @@ export default Ember.Component.extend(ThrottledResize, {
     var socket = this.get('socket');
     if (term && socket) {
       var geometry = proposeGeometry(term);
-      socket.send("4" + window.btoa(`{"Width":${geometry.cols},"Height":${geometry.rows}}`));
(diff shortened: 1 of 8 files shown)


loganhz:
You need to use `rancher/server:master` to test it. But it looks like you can not go to project page now as `namespace` is not under project schema in this build.

lib/shared/addon/components/container-shell/component.js
```diff
@@ -90,26 +85,30 @@ export default Ember.Component.extend(ThrottledResize, {
       });
       this.set('term', term);
 
-      term.on('data', function(data) {
-        //console.log('To Server:',data);
-        socket.send(AWS.util.base64.encode(data));
+      term.on('data', function (data) {
+        socket.send(`0${window.btoa(window.unescape(encodeURIComponent(data)))}`);
```
vincent99: Is there a reason for this hackery (and in the other places) instead of `AWS.util.base64.encode()` ?  `window.btoa` doesn't work with arbitrary binary data, the AWS one does.

loganhz: Updated

loganhz:
It looks like the `exec` feature is broken in `rancher/server:master`. `kubectl exec` doesn't work.

PR #1659, opened 2018-02-24
Alert: Fixed sorting & searching and filtering bugs.

vimniky:
Fixed:

https://github.com/rancher/rancher/issues/11605
https://github.com/rancher/rancher/issues/11597
https://github.com/rancher/rancher/issues/11596
https://github.com/rancher/rancher/issues/11576
https://github.com/rancher/rancher/issues/11575
https://github.com/rancher/rancher/issues/11574

### DIFF
--- a/lib/alert/addon/components/alert-table/component.js
+++ b/lib/alert/addon/components/alert-table/component.js
@@ -21,8 +21,8 @@ const headers = [
   {
     translationKey: 'alertPage.index.table.target',
     name: 'target',
-    searchField: ['target', 'displayTargetType'],
-    sort: ['target', 'id'],
+    searchField: ['nodeName', 'resourceKind', 'displayTargetType'],
+    sort: ['nodeName', 'id', 'resourceKind'],
   },
   {
     translationKey: 'alertPage.index.table.condition',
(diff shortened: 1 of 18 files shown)


lib/alert/addon/components/alert-table/component.js
```diff
@@ -54,7 +57,30 @@ export default Component.extend({
   search: true,
   searchText: null,
 
+  clusterAlerts: function() {
+    const clusterId = get(this, 'clusterId');
+    const gs = get(this, 'globalStore');
+    return gs.all('clusterAlert').filterBy('clusterId', clusterId);
+  }.property('clusterId'),
```
vincent99: This needs to depend on `clusterAlert.@each.clusterId`, it's now not a live list and needs to be recomputed if a cluster comes or goes.  similar below.

vimniky: Thanks! PR Updated.

vimniky: You mean I need to change the code like the following? I don't know that clusterAlerts can depend on itself. 

```
  clusterAlerts: function() {
    const clusterId = get(this, 'clusterId');
    const gs = get(this, 'globalStore');
    return gs.all('clusterAlert').filterBy('clusterId', clusterId);
  }.property('clusterId', 'clusterAlerts.@each.{clusterId}'),
```

PR #1661, opened 2018-02-26
workload port

loganhz:


loganhz:
It looks like the backend only supports `containerPort ` and `protocol ` of `workload.containers.ports`. And there is no `publicEndpoints` in the API.
So I can not test the function.

PR #1715, opened 2018-03-07
Bugfix: wrong cluster node count

vimniky:
Related issue:
https://github.com/rancher/rancher/issues/11152


vincent99:
I don't think this makes sense.  The count is how many nodes there are, their state is irrelevant.  https://github.com/rancher/rancher/issues/11152#issuecomment-371017801

PR #1718, opened 2018-03-07
Update searchable-component search algorithm

vimniky:
Related issue:
https://github.com/rancher/rancher/issues/12003

### DIFF
--- a/lib/shared/addon/components/searchable-select/component.js
+++ b/lib/shared/addon/components/searchable-select/component.js
@@ -172,23 +172,40 @@ export default Component.extend({
     }
 
     if ( this.get('clientSideFiltering') ) {
-      return options.filter(option => {
-        const segments = filter.split(' ');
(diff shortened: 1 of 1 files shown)


vincent99:
- This has conflicts from changes for input-identity last night.
- I don't think "starts with" is going to be helpful because of things like if you type "Volume" you'd probably like to see "PersistentVolumeClaims".  
- Splitting on space makes no sense anymore, no entry will start with both terms.

@vimniky If you want to improve this, do both:
  - Find results that start with one of the search terms (and contain all) and return them at the top of the list
  - Then add additional results that contain all terms (but did not start with one).
  - Continue using toLowerCase(), startsWith(), and includes()... Regular expressions and creating one every loop is relatively slow

vimniky:
@vincent99  I Updated the PR to what you suggested.

PR #1748, opened 2018-03-22
Pipeline fix 18/3

biblesyme:
Fix issue:

1. Add Pull Request Webhook 
2. Fix issue rancher/rancher#11675
3. Enable user to use the public repository to create Pipeline 
4. Set `access.principal` instead of overriding `access.me`


zionwu:
https://github.com/rancher/rancher/issues/11678

PR #1757, opened 2018-03-23
row check

lvuch:


vincent99:
I broke this and fixed a different way elsewhere..

PR #1759, opened 2018-03-24
Fix the issue that user can not see Authentication in nav

loganhz:


vincent99:
Why is this needed?  The nav menu should be based on rbac-allows PUT to authConfigs or something like that.

loganhz:
image

I'm trying to fix https://github.com/rancher/rancher/issues/12227 in current version.

I am using `docker run -d -p 80:80 -p 443:443 rancher/server:master` to start rancher server on my laptop.

Both of admin user and standard user have the `PUT` to `authConfigs`.

PR #1781, opened 2018-04-02
Add custom log path

vimniky:
Related issue:

https://github.com/rancher/rancher/issues/11625

### DIFF
--- a/lib/shared/addon/components/container/form-volumes/component.js
+++ b/lib/shared/addon/components/container/form-volumes/component.js
@@ -1,6 +1,7 @@
 import { all } from 'rsvp';
 import { inject as service } from '@ember/service';
 import { get, set } from '@ember/object';
+import { alias } from '@ember/object/computed';
 import Component from '@ember/component';
 import layout from './template';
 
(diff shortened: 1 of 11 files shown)


vincent99:
This should go under Volumes -> Add Volume 

image

and work as a normal `lib/shared/addon/components/volume-source/source-*`, similar to the longhorn one.

vimniky:
PR updated, and now: the custom log path configuration goes into the volumes section and will only shows up if cluster or project logging is enabled.

lib/shared/addon/components/container/form-volumes/component.js
```diff
@@ -11,12 +12,17 @@ export const EXISTING_PVC = 'existingPvc';
 export const BIND_MOUNT = 'bindMount';
 export const TMPFS = 'tmpfs';
 export const SECRET = 'secret';
+export const CUSTOM_LOG_PATH = 'customLogPath';
+
+export const LOCAL_FLEXVOLUME = 'rancher.io/localflexvolume'
```
vincent99: Did this get renamed?

vimniky: No, I added that.

vimniky: I need to distinguish local flex volume from flex volume according to the `volume.flexvolume.options.dirver` field which either  `"rancher.io/localflexvolume"` or `"rancher.io/flexvolume"` .

vincent99: I mean I thought we were renaming the flex driver to `log-aggregator` or something, because "localflexvolume" conveys no meaning about what it does.

PR #1802, opened 2018-04-05
Fix %MISSING% text in the change password button

alex-kovalchuk:
In the password change popup, the confirmation button with the text %MISSING%. I tried to fix it.


loganhz:
link to https://github.com/rancher/rancher/issues/12540

PR #1818, opened 2018-04-12
Fix

loganhz:
rancher/rancher#12432
rancher/rancher#12466
rancher/rancher#12002
rancher/rancher#12381
rancher/rancher#12639
rancher/rancher#12207
rancher/rancher#12544
rancher/rancher#12435
rancher/rancher#11936
rancher/rancher#11986
rancher/rancher#11649
rancher/rancher#10980
rancher/rancher#11780
rancher/rancher#12517
rancher/rancher#12513
rancher/rancher#12643
rancher/rancher#12607


app/models/node.js
```diff
@@ -122,16 +122,11 @@ var Node = Resource.extend(StateCounts, ResourceUsage, {
     const roles = get(this, 'rolesArray');
 
     if ( roles.length >= 3 ) {
-      return [intl.t('generic.all')];
```
vincent99: I think this just needed to watch `intl.locale`, which I added separately... If the roles are not the actual translated strings then searching by role becomes weird/broken.

app/models/pod.js
```diff
@@ -32,7 +33,7 @@ var Pod = Resource.extend(DisplayImage, {
     shell() {
       get(this, 'modalService').toggleModal('modal-shell', {
         model: this,
-        escToClose: false,
+        escToClose: true,
```
vincent99: Hitting escape inside the terminal should go to the terminal, for things like... anything in vim...

lib/shared/addon/components/container/form-volumes/component.js
```diff
@@ -37,25 +37,25 @@ export default Component.extend({
...
-    (get(this, 'workload.volumes')||[]).forEach((volume) => {
-      if ( volume.persistentVolumeClaim ) {
+    (get(this, 'workload.volumes') || []).forEach((volume) => {
+      if (volume.persistentVolumeClaim) {
         entry = {
           mode: EXISTING_PVC,
           volume,
         };
-      } else if ( volume.hostPath ) {
+      } else if (volume.hostPath) {
```
vincent99: Do you have some tooling that's making spacing changes?  This is an an odd combination of adding and removing them througout this file.   Spaces are cheap and the standard in this repo is generally more is better
```
if ( condition ) {
  fn(arg1, arg2)
}
<probably a blank line before more logic>
```

Don't go back and undo it all now, but not touching existing ones in the middle of a large PR is preferable.  If you want to bulk cleanup, make it a separate PR.

lib/shared/addon/components/form-ingress-backends/component.js
```diff
@@ -43,7 +43,7 @@ export default Component.extend({
         pathArray.pushObject(get(this, 'store').createRecord({
           type: 'httpingresspath',
           backendType: 'service',
-          targetPort: get(path, 'targetPort'),
+          targetPort: get(path, 'targetPort').toString(),
```
vincent99: This will blow up if targetPort is null;  `(get(path, 'targetPort')||'')+''`

lib/shared/addon/components/form-ingress-backends/component.js
```diff
@@ -43,7 +43,7 @@ export default Component.extend({
...
           serviceId: get(path, 'serviceId').replace('/', ':'),
```
vincent99: This is not new and I don't know what it's doing, but it's probably very wrong.  If the UI is still generating `$stack/$service` name that should be corrected to return the right ID rather than trying to string replace into a valid one.

lib/shared/addon/components/new-select/component.js
```diff
@@ -74,9 +75,11 @@ export default Component.extend({
   },
 
   // dont know why this is here but it causes issues trying to set on undefined objects
-  // didRender() {
-  //   this._change();
-  // },
+  didRender() {
+    if (this.get('setValueOnInit')) {
```
vincent99: What issue was this change for?  This seems like it's going to break anywhere that there's a value that's no longer in the list of choices.

PR #1820, opened 2018-04-16
Bug fix & Enhancement

biblesyme:
Add action `Edit As Yaml `
Add a Version column to Roles and before CPU
Fix issue rancher/rancher#12676
Fix issue rancher/rancher#12670


biblesyme:
@vincent99 Ok, I will change to a Modal. Using store.request works now, thanks. 
But, I find that I can change the name of the resource if I edit it as YAML, it returns `400` with `the name of the object (test4) does not match the name on the URL (test)`

vincent99:
The actual name is not normally changeable in k8s; `name` is the `id` in out API, and what we show as name is `displayName`.

PR #1831, opened 2018-04-18
Fix app references

vincent99:


loganhz:
thanks for your update! I will take care next time

PR #1842, opened 2018-04-23
Prevent notifier from being deleted when there are alerts referencing it

vimniky:
Related issue: https://github.com/rancher/rancher/issues/12914

### DIFF
--- a/app/models/notifier.js
+++ b/app/models/notifier.js
@@ -1,10 +1,14 @@
 import Resource from 'ember-api-store/models/resource';
 import { inject as service } from '@ember/service';
-import { get } from '@ember/object';
+import { computed, get } from '@ember/object';
+import { hash, reject } from 'rsvp';
 
 export default Resource.extend({
   type: 'notifier',
+  growl: service(),
+  intl: service(),
 
+  globalStore: service(),
   modalService: service('modal'),
 
   init(...args) {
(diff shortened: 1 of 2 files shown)


vincent99:
I'd rather see notifier's `canRemove` property be computed to properly reflect whether it's allowed to be deleted or not.  Eliminating bulk delete is not great, and this doesn't _actually_ prevent deleting.  If you do cmd+click delete it's just going to go through, because all your logic is in `promptDelete()`.

You should be able to add `hasMany`'s in notifier to find the project and cluster alerts that are using the notifier, then compute canRemove from the length of those.

vimniky:
I have to use hasMany and the looks like the following:

```js
  clusterAlerts: hasMany('id', 'clusterAlert', 'recipients.@each.recipientId', 'globalStore'),
  projectAlerts: hasMany('id', 'projectAlert', 'recipients.@each.recipientId', 'globalStore'),
```
But turns out it didn't works.  
The `alert.recipients` contains an array of notifiers, not just a normal filed.  In this situation can I still use hasMany?

vincent99:
No, it doesn't handle `@each` / arrays.  That's not trivial to add, so I guess go with 

> Alternately, allow delete in availableActions but override delete action in the model to throw a growl and not do anything if there are alerts using it.

That still allows bulk delete, and can tell you why you can't delete a specific entry.

vimniky:
@vincent99  I have some issue with enabling bulk deletion. I can't override the delete bulk action handler in my model because bulk actions are handled in sortable-table.  Another thing is that I can't simply set `canRemove`  to `false` to disable deletion because we want to let the users know what it can't be deleted and what alerts are using this notifier.

vincent99:
You're not supposed to be able to easily override it; that's the point of them being centralized.

But you don't need to.  I'm talking about overriding the actual `delete` action that performs the delete to decide to do nothing if it's in use.

```javascript
#.../models/notifier.js
{
  growl: service(),

  actions: {
    delete() {
      const alertsUsingMe = this.notifiersUsingMe();
      if ( alertsUsingMe.length ) {
        get(this, 'growl').error(... "This notifier is in use by: " + alertsUsingMe.join(","));
      } else {
        return this._super(...arguments);
      }
    }
  },

  alertsUsingMe: computed(function() {
    ...find array of alerts using this notifier
  }).volatile(), // Will recompute every time it's asked for, since observing nested fields is a problem.
}

PR #1858, opened 2018-04-27
Fix

loganhz:
rancher/rancher#13075
rancher/rancher#13119
rancher/rancher#12685
rancher/rancher#13089
rancher/rancher#13104


loganhz:
image

Hi, @vincent99 

@superseb wants to see the os information in node list page. 
Can we do this? maybe dropping it a line would probably take too much space.

PR #1871, opened 2018-05-02
Fix custom node label issue

loganhz:
https://github.com/rancher/rancher/issues/12691


vincent99:
https://github.com/rancher/rancher/issues/12691#issuecomment-386411132

> What? No.. these are labels to be given to the agent container to then add on to the node, just like in 1.x. They are not labels to assign to the agent container itself, why would anybody want that?

PR #1874, opened 2018-05-02
Support max and min for catalog int input

loganhz:


vincent99:
Seems like both `schema/input-int` and `schema/input-float` are redundant;  they extend `lib/shared/addon/components/input-integer` and `.../input-number` instead of you duplicating all the code.

Or eliminate them entirely, but I think in dynamic schema things like catalog answers it's assumed that `schema/input-{type}` exists for any valid API type.

loganhz:
@vincent99 updated. Thanks

PR #1880, opened 2018-05-04
Fix

n313893254:
https://github.com/rancher/rancher/issues/13195
https://github.com/rancher/rancher/issues/13114
https://github.com/rancher/rancher/issues/13287


lib/shared/addon/utils/util.js
```diff
@@ -345,6 +345,26 @@ export function isBadTld(name) {
   }
 }
 
+export function formatEndpoint(endpoint) {
```
vincent99: Can you use the parser in lib/shared/addon/utils/parse-uri.js instead of adding another one?

n313893254:
sure

n313893254:
updated

n313893254:
updated

PR #1884, opened 2018-05-06
Preserve route

loganhz:
https://github.com/rancher/rancher/issues/8486


app/authenticated/project/route.js
```diff
@@ -52,6 +59,13 @@ export default Route.extend(Preload,{
     get(this, 'scope').finishSwitchToProject(get(model,'project'));
   },
 
+  redirect() {
```
vincent99: This needs to only be in the index route (which already had it, so why here?) (and same for cluster)... all sub-routes of project go through here and you can end up in a situation where you tried to load `/p/<id>/workspaces` but the setting says `members` or whatever and you ignore where the URL says to go on pageload.

loganhz: I tried to add them in the index route.

But I don't know if user clicks `workloads` on the menu or switches to the project from another project.

The `redirect` method in the index route will be triggered in these two cases.

PR #1885, opened 2018-05-07
Enhance catalog answer validation

loganhz:


app/models/templateversion.js
```diff
@@ -1,33 +1,91 @@
...
 import Resource from 'ember-api-store/models/resource';
+import { get, computed } from '@ember/object';
+import { evaluate } from 'shared/utils/evaluate';
 import C from 'ui/utils/constants';
 
 export default Resource.extend({
   scope: service(),
 
-  headers: function() {
+  validationErrors() {
```
vincent99: This is all duplicate logic from  https://github.com/rancher/ember-api-store/blob/4d519950a0749765a0a405eb892bc6c74aa7eed4/addon/models/resource.js#L54-L157 and https://github.com/rancher/ember-api-store/blob/master/addon/utils/validate.js , minus translation support... De-duplicate or refactor if they need to be different

loganhz:
Done.
Now using  `validateChars` and `validateLength`  from `ember-api-store/utils/validate`

PR #1890, opened 2018-05-08
Update Aliyun ECS node driver UI

loganhz:


vincent99:
Can you make API calls to them through the proxy to validate API credentials and lookup instance types and other stuff (like AWS or DO, except for some parts AWS doesn't have the info in the API) instead of maintaining lib/shared/addon/utils/aliyunecs-choices.js ?  It's a long feedback loop for creating a node template and a pool and it fails and go back and edit and try again so I'd like all the drivers to actually validate things and talk to the API beforehand...

loganhz:
Done. Now it makes API calls to Aliyun ECS through the proxy like AWS.

PR #1902, opened 2018-05-12
Support sidebar menu

loganhz:


vincent99:
@lvuch can you wireframe/tweak this; my comments:

- Moving the stuff in the footer to the header seems weird to me.  If you're going to do left nav, there should be no header.
- User menu stuff should be with the user, I don't understand why it's half a dropdown with just the user info and logout while the other options are in the left nav.
- The user menu could be attached to the bottom-left corner of the nav so that it's still separated.
- Cluster dropdown has to be on top of or inside the left nav
- Hamburger should be before the cow, I think
- `overflow: auto`, not `scroll`.

visualapps:
hey @loganhz it would be good if u attach the visual design of sidebar menu, It will help to visualise the changes on design.

vincent99:
Going a different direction with this...

PR #1936, opened 2018-05-30
Fix

loganhz:
rancher/rancher#13550
rancher/rancher#13523
rancher/rancher#13690


lib/shared/addon/mixins/node-driver.js
```diff
@@ -92,6 +93,16 @@ export default Mixin.create(NewOrEdit, ManageLabels, {
     this._super(...arguments);
 
     if ( !get(this, 'editing') && typeof get(this, 'bootstrap') === 'function') {
```
vincent99: Why can't this be in `bootstrap()` of the aliyun driver instead of in the middle of generic driver?

loganhz: @vincent99 yes, it can be. I have moved it into aliyun driver. Thanks

PR #1937, opened 2018-05-30
Catalog fix

loganhz:
rancher/rancher#13519
rancher/rancher#13692
rancher/rancher#13693

### DIFF
--- a/app/models/service.js
+++ b/app/models/service.js
@@ -1,9 +1,11 @@
 import Resource from 'ember-api-store/models/resource';
 import { get, computed } from '@ember/object';
 import { inject as service } from '@ember/service';
+import Util from 'ui/utils/util';
 
 export default Resource.extend({
   intl: service(),
+  scope: service(),
 
   canEditYaml: true,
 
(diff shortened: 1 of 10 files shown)


app/models/service.js
```diff
@@ -16,4 +17,21 @@ export default Resource.extend({
       return intl.t('model.service.displayKind.generic');
     }
   }),
+
+  proxyEndpoints: computed('labels', function(){
+    const parts = []
+    const labels = get(this, 'labels');
+    if ( labels && labels['kubernetes.io/cluster-service'] === 'true' ) {
+      (get(this, 'ports') || []).forEach((port) => {
+        const linkEndpoint = `${location.protocol}//${location.host}/k8s/clusters/${get(this, 'scope.currentCluster.id')}/api/v1/namespaces/${get(this, 'namespaceId')}/services/${get(port, 'targetPort')}:${get(this, 'name')}:/proxy/`;
```
vincent99: I'm surprised linting passes with this, it probably just doesn't support parsing the backtick string interpolation expressions.  `location` looks like an undefined variable, always use `window.` for things that are intentionally referring to global window variables.

Also `${location.protocol}//${location.host}` is just `${window.location.origin}`

PR #1939, opened 2018-05-31
fix link for port requirements

deniseschannon:


deniseschannon:
https://github.com/rancher/rancher/issues/12630

PR #1949, opened 2018-06-07
Fix

loganhz:
https://github.com/rancher/rancher/issues/13708


vincent99:
btoa/atob does not handle arbitrary binary.  The AWS base64 does.

loganhz:
@vincent99 Updated

PR #1960, opened 2018-06-11
Support logging SSL for splunk, ES and kafka

n313893254:
Related issue:
https://github.com/rancher/rancher/issues/13526
https://github.com/rancher/rancher/issues/13065

### DIFF
--- a/lib/logging/addon/components/logging/form-log-docker/component.js
+++ b/lib/logging/addon/components/logging/form-log-docker/component.js
@@ -1,0 +1,4 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+});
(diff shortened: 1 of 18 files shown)


aiwantaozi:
https://github.com/rancher/rancher/issues/13755

vincent99:
You have basically the same thing over and over.  Make a component instead of copying and pasting and compiling the same ssl config repeatedly.  Every byte of template you put in is several bytes every user has to download.

And there's already an advanced-section component.

n313893254:
Thank you for review, I will modify the code.

n313893254:
@vincent99 Fixed

PR #1966, opened 2018-06-13
Do not create namespace for app to kube-system

loganhz:
https://github.com/rancher/rancher/issues/13519
https://github.com/rancher/rancher/issues/13958


loganhz:
https://get.docker.com/ keeps return 403. No build can pass

PR #1971, opened 2018-06-15
Fix remove pod issue and sort display endpoints

loganhz:
rancher/rancher#12498
rancher/rancher#13694

### DIFF
--- a/app/container/template.hbs
+++ b/app/container/template.hbs
@@ -6,9 +6,7 @@
   </div>
   <div class="right-buttons">
     {{badge-state model=model}}
-    {{#unless (eq model.state 'removed')}}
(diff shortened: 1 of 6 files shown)


app/container/template.hbs
```diff
@@ -6,7 +6,9 @@
   </div>
   <div class="right-buttons">
     {{badge-state model=model}}
-    {{action-menu model=model showPrimary=false classNames="ml-10 inline-block" size="sm"}}
```
vincent99: Why?  There's still actions on removed things, like view in api.  If there's wrong things in the menu then the availableActions is wrong for the resource.

loganhz:
@vincent99 Updated. Thanks

loganhz:
Go back to the list after the pod state is removed.

PR #1994, opened 2018-06-22
[Bug Fix]continuous click disable helm button causes multiple requests to be sent

ly5156:


lib/global-admin/addon/catalog/controller.js
```diff
@@ -25,7 +25,7 @@ export default Controller.extend({
         get(this, 'growl').fromError('Error removing Library', err);
       }).finally(() => {
         set(this, 'togglingLibrary', false);
-        this.send(this, 'refresh');
+        this.send('refresh', this);
```
vincent99: These should all just be `this.send('refresh');`, the action in the route doesn't take an argument.

lib/global-admin/addon/catalog/template.hbs
```diff
@@ -14,7 +14,7 @@
     <div>
       {{#if library}}
         <div class="btn-group no-inline-space">
-          <button class="btn btn-link btn-sm bg-default" {{action "disableLibrary"}} disabled={{togglingLibrary}}>{{#if togglingLibrary}}<i class="icon icon-spinner icon-spin"></i> {{/if}}{{t 'generic.disable'}}</button>
+          <button class="btn btn-link btn-sm bg-default" {{action "disableLibrary"}} disabled={{disabledLibrary}}>{{#if togglingLibrary}}<i class="icon icon-spinner icon-spin"></i> {{/if}}{{t 'generic.disable'}}</button>
```
vincent99: Add `waitForTransition()` after the save calls in the disable actions (`[...].save().waitForTransition().[...]`), instead of adding new disabled computed properties.  Then `toggling[thing]` will just stay true until the remove is complete, and you'll disable the button and show the spinner instead of showing a disabled button that has no explanation and no indication of activity.

lib/global-admin/addon/components/modal-edit-catalog/component.js
```diff
@@ -25,7 +33,23 @@ export default Component.extend(ModalBase, NewOrEdit, {
     set(clone, 'kind', 'helm');
     set(this, 'model', clone);
   },
+  validate: function() {
+    const intl = get(this, 'intl');
+    var model = get(this, 'primaryResource');
+    var name = get(this, 'primaryResource.name')
+    var errors = model.validationErrors();
+    if (name && existingCatalogs.indexOf(name.toLowerCase()) > -1) {
```
vincent99: `existingCatalogs.includes(name.toLowerCase())`.  But this logic is wrong in general... you can't hardcode existingCatalogs, it should check against the ones that actually exist right now.

If the standard `library` is disabled then you should to be able to add your own catalog with the name library and an arbitrary URL.

And if I already have a catalog named 'foo', you might as well not let me add a second one named that since the API is going to say no anyway.

translations/en-us.yaml
```diff
@@ -1827,7 +1827,8 @@ catalogSettings:
     branch:
       label: Branch
       placeholder: 'e.g. master'
-
+  errors:
+    nameInExists: '{key} is reserverd. Please input a different name'
```
vincent99: `There is already a catalog named `{key}`.  Please choose a different name.`

translations/zh-hans.yaml
```diff
@@ -1832,7 +1832,8 @@ catalogSettings:
     branch:
       label: 分支
       placeholder: '例如：master'
-
+  errors:
+    nameInExists: '{key} 为系统预留名称，请使用新的名称'
```
vincent99: The equivalent to the english change --^

PR #1995, opened 2018-06-22
count the right number

StrongMonkey:


StrongMonkey:
roles should four instead of three

PR #2003, opened 2018-06-27
Fix log and shell issue

loganhz:
User can not input `%` now

### DIFF
--- a/app/components/container-logs/component.js
+++ b/app/components/container-logs/component.js
@@ -91,7 +91,7 @@ export default Component.extend({
       let ansiup = new AnsiUp.default;
       set(this, 'status', 'connected');
       var isFollow = ($body.scrollTop() + $body.outerHeight() + 10) >= body.scrollHeight;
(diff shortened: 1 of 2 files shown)


app/components/container-logs/component.js
```diff
@@ -91,7 +91,7 @@ export default Component.extend({
       let ansiup = new AnsiUp.default;
       set(this, 'status', 'connected');
       var isFollow = ($body.scrollTop() + $body.outerHeight() + 10) >= body.scrollHeight;
-      const data = decodeURIComponent(AWS.util.base64.decode(message.data));
```
vincent99: AFAIR the 1.x shell used to encode the data on the server-side;  Since the k8s server isn't, shouldn't it just be removed?  decode(encode()) sounds like no-op.

loganhz: Yes! You are right. Thanks

loganhz:
@vincent99 Updated. Removed the unused `decode(encode())`

PR #2040, opened 2018-07-05
Fixed misprints and grammar in russian translate

zlyoha:
Changed Pods translate to "Капсулы"
Added comma (according to the rules of grammar)
Fixed misprints


vincent99:
Hi; these need to go into http://translate.rancher.com/ , CrowdIn is the "source of truth" for all languages except English and Chinese that we maintain in the source tree.  The rest are periodically synched from CrowdIn and any direct changes to them in this repo will be overwritten.

PR #2048, opened 2018-07-09
Wrap log lines

loganhz:
https://github.com/rancher/rancher/issues/13636


vincent99:
This should be optional with a "Wrap lines" checkbox in the log screen, and remember the state as a user setting (which can default to on/wrap).  

Wrapping is very unhelpful in some situations like when you're watching wide lines that vary slightly at the end but you don't care much about that end part, e.g. web server access logs.

loganhz:
@vincent99 Updated. Thanks for your review

vincent99:
Setting values are automatically JSON encoded/decoded by the service... So you can use regular `true`/`false`.

Otherwise LGTM

loganhz:
@vincent99 Updated

PR #2058, opened 2018-07-16
Change used to reserved for CPU and memory gauges

superseb:
With all the questions on the dashboard, it's better to change to reserved for CPU and Memory. If you agree, we need to add translations for Chinese before merge.


vincent99:
@superseb Yes that's fine

loganhz:
@superseb  Created another PR includes your changes. https://github.com/rancher/ui/pull/2062
We need to update translations for all languages which have `clusterDashboard. subtitle`.

In that PR, only `en-us` and `zh-hans` show `reserved` for CPU and memory gauges. Others still show `used`. But the format is changed.

superseb:
Thanks!

PR #2061, opened 2018-07-17
catalog certification

lvuch:


loganhz:
related issue https://github.com/rancher/rancher/issues/14574

guangbochen:
Can we please also add `experimental` label to the catalog certification, need it for catalog apps like longhorn and nfs-provisioner, thanks.
```
io.rancher.certified: Experimental
```

vincent99:
`rancher` and `partner` show text from a translation and special colors; any other arbitrary value shows the text in a third color.  So you can put in experimental if you want.

vincent99:
image

guangbochen:
thanks, it works but only the UI seems slightly different the above post.
And it only works for Rancher Library catalog correct?
image

vincent99:
That's the old screen, I haven't updated the latest2 build yet.

PR #2062, opened 2018-07-18
Change subtitle for gauges

loganhz:


vincent99:
The translations other than zh-hans and en-us need to go into CrowdIn or they're going to get lost next time they are synched..

PR #2066, opened 2018-07-19
Fix catalog clone issue

loganhz:
https://github.com/rancher/rancher/issues/14629


guangbochen:
This build seems failed and we still encounter the `displayName not defined` issue with catalog, thanks.

PR #2083, opened 2018-07-23
Project level resource quota

loganhz:
https://github.com/rancher/rancher/issues/13277
https://github.com/rancher/rancher/issues/14620

### DIFF
--- a/app/authenticated/cluster/projects/new-ns/controller.js
+++ b/app/authenticated/cluster/projects/new-ns/controller.js
@@ -1,7 +1,7 @@
 import Controller from '@ember/controller'
 import NewOrEdit from 'ui/mixins/new-or-edit';
 import { alias } from '@ember/object/computed';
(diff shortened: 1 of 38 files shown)


vincent99:
This looks good but is missing one piece;  On a namespace create/edit you need to be able to choose which quota template is used (`resourceQuotaTemplateId`).    That should probably also be a column in the namespace table.

loganhz:
@vincent99 Done. Now user can only select `resourceQuotaTemplateId` when create/edit a namespace under a project which has `resourceQuota`  field.  
Alena said we would allow user to select `resourceQuotaTemplateId` for all namespaces later.

vincent99:
One more thing, can you make memory fields take `MiB`, cpu `milli CPUs` like the workload definition do (and storage fields in 10^9 `GB`):

image

So that the user can just type in a number instead of having to put in a unit.

loganhz:
@vincent99 Done

PR #2086, opened 2018-07-24
Rancher Logging does not take custom Docker root dir into account

n313893254:
https://github.com/rancher/rancher/issues/13304

### DIFF
--- a/lib/global-admin/addon/clusters/new/route.js
+++ b/lib/global-admin/addon/clusters/new/route.js
@@ -14,6 +14,9 @@ export default Route.extend({
     let globalStore = this.get('globalStore');
 
     let cluster = globalStore.createRecord({ type: 'cluster' });
+    let defaultValue = { dockerRootDir: cluster.dockerRootDir, }
+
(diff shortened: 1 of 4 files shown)


vincent99:
Is there a backend PR for this somewhere?  `dockerRootDir` and `defaultValue[.dockerRootDir]` are not public fields of a `cluster`.

n313893254:
`dockerRootDir,` Backend had merged https://github.com/rancher/types/pull/505/commits/17506231edfbfe436dbec99dc14dcd15d00961a7
But `defaultValue[.dockerRootDir]` is just used for display a placeholder value.  It can also store in another field in `model,` what do you think?

vincent99:
Ok I see `dockerRootDir` on the model now.. but why are you taking it and adding it to a new fake property on the cluster model that will never be saved, just to show it?  

Instead of just directly using `(t 'clusterNew.rke.dockerRootDir.placeholder' ... dir=model.cluster.dockerRootDir)`?

n313893254:
You are right, I fixed it and obtained the default value by creating a record in component init.

PR #2096, opened 2018-07-26
Keep virtual center and cloud provider in cluster yaml consistent with RKE

loganhz:
https://github.com/rancher/rancher/issues/14750


vincent99:
No... Random special keys don't go in general purpose util functions.

Fix them in the specific place it's a problem and/or add a generic way to exclude keys from being removed if empty.

(also removeEmpty is already very hard to read)

loganhz:
Done. Updated

PR #2126, opened 2018-08-01
Coerce catalog item version before compare

loganhz:
https://github.com/rancher/rancher/issues/14861


vincent99:
This seems to just mask a symptom instead of addressing the problem. Why does deploying that catalog item result in an app that doesn't correlate back to the version that it was deployed from??

PR #2140, opened 2018-08-06
Support custom catalog answer without question defined

loganhz:


superseb:
https://github.com/rancher/rancher/issues/14964

PR #2145, opened 2018-08-08
Add node name in advacned options when adding custom cluster

loganhz:
https://github.com/rancher/rancher/issues/13999


FinlayDaG33k:
It would be neat if it would suggest a random node name (like Docker does with container names) if no `--node-name` is specified.

PR #2149, opened 2018-08-11
Fix waiting for namespace creation

vincent99:


cjellick:
https://github.com/rancher/rancher/issues/13703

PR #2171, opened 2018-08-22
Use lower case for name

loganhz:
https://github.com/rancher/rancher/issues/15013


vincent99:
Not everything requires lowercase names.  Things that do should be `type: dnsLabel` and get automatically lowercased during validation.

loganhz:
Updated. Thanks

PR #2184, opened 2018-08-24
Browser gets stuck while Viewing huge logs using rancher's view logs operation

ly5156:
https://github.com/rancher/rancher/issues/14535


loganhz:
@ly5156 Can you close this PR, please? We can fix it in https://github.com/rancher/ui/pull/2186

PR #2235, opened 2018-09-10
Update translations from crowdin

vincent99:
@loganhz, can you look at zh-hant and why there's so many lines removed?  Not clear to me if they are things that were never traditional, were intentionally removed, were the same as simplified, were mistakenly removed, or what.

### DIFF
--- a/app/components/page-footer/template.hbs
+++ b/app/components/page-footer/template.hbs
@@ -4,14 +4,14 @@
   {{#if projectId}}
     <a href="{{href-to 'authenticated.project.help' projectId}}" role="button" class="btn btn-sm bg-transparent">{{t 'pageFooter.help'}}</a>
   {{/if}}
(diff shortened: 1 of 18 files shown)


loganhz:
@vincent99 
@xiaoluhong updated the both zh-hans and zh-hant in code directly. I guess we should keep the current zh-hant in our repo.

xiaoluhong:
@vincent99 Hi, I translated it directly through the repo, not in the crowded translation，Basically keep the Chinese and English synchronized.

vincent99:
Oh ok, I'll pull it out then.  You should be periodically pushing those to crowdin, or we need to remove it as a language there and only take PRs.

xiaoluhong:
We choose zh-hans and zh-hant to submit it through PR, which will be more convenient. Thank you.

PR #2332, opened 2018-10-11
Fix the issue that UI loads vendor css twice

loganhz:
Proposed changes
======
Currently Rancher UI will load `vendor.css` twice.

image

Types of changes
======

Linked Issues
======
 

Further comments
======


loganhz:
> Adding reference to the UI issue
> 
> [rancher/rancher#16084](https://github.com/rancher/rancher/issues/16084)
> 
> we must have an issue created from now on so that QA has ways to track and test fixes.

Got it! Thanks

PR #2337, opened 2018-10-13
Fix engine dependency issue

loganhz:
Proposed changes
======
Fix nodes engine dependency issue. `intl` is used in aliyun ecs driver UI.

Also removed the duplicated entries in pipeline dependency

Types of changes
======
Bugfix

Linked Issues
======
https://github.com/rancher/rancher/issues/16121

 

Further comments
======


loganhz:
There is another issue https://github.com/rancher/rancher/issues/16122

@westlywright

PR #2356, opened 2018-10-23
Use k8s resource names in UI

superseb:
Proposed changes
======
Use k8s resource names as is.

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/16247

Further comments
======

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -217,9 +217,9 @@ appDetailPage:
     detail: Persistent Volume claims created with this application
     nodata: No volume claims were made for this application.
   configMaps:
-    title: Config Maps
-    detail: Config Maps associated with this application
-    nodata: This application has no config maps
+    title: ConfigMaps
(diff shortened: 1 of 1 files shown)


vincent99:
No?

The UI is for humans.  The type names can't have spaces for meaningless technical reasons, that doesn't mean we need to propagate that into the UI.  Similarly, we don't display exactly the field name every control is manipulating, or every field, or necessarily a 1:1 mapping at all.

But if we were going to change this, this addresses only a random subset of them and introduces new inconsistencies (like "Use a secret" and "Use a certificate" remaining lowercase right next to "Use a ConfigMap").  And there are other more relevant examples that are actually legitimately confusing to someone who knows the k8s terms (like the way Volume, PV, and PVC are used).

superseb:
Fine by me, I just see terminology that's not 1-on-1 with k8s terminology/docs. If I don't file it, someone else will and now we know why it's like this.

PR #2396, opened 2018-11-14
Fix service port ingress issue

loganhz:
Proposed changes
======
If users use a service as target backend for ingress, UI allows users to input any port.
But the ingress will not work if the port is not exposed by the service.

So we need to change the port input to a dropdown and show a warning to explain it. 
Users are only able to select the available ports in a service.

Types of changes
======
Bugfix (non-breaking change which fixes an issue)

Linked Issues
======
https://github.com/rancher/rancher/issues/15809

 

Further comments
======


lib/shared/addon/components/form-ingress-backends/template.hbs
```diff
@@ -65,14 +68,19 @@
 
           <td class="p-5" data-title="{{t 'formIngressBackends.targetPort.label'}}">
             {{#input-or-display editable=editing value=path.targetPort classesForDisplay="clip"}}
-              {{#if (and (eq path.backendType 'service') (eq path.service.recordType 'workload'))}}
+              {{#if (and path.service (eq path.backendType 'service'))}}
```
vincent99: Avoid repeating conditions in multiple ifs..

```
{{#if (eq path.backendType 'service')}}
  {{#if path.service}}
    // first case
  {{else}}
    // second case
  {{//if}}
{{else}}
  // third case
{{/if}}```

File: lib/shared/addon/components/form-ingress-backends/template.hbs [hunk 2]
```diff
@@ -65,14 +68,19 @@
...
                 {{new-select
                   classNames="form-control input-sm"
                   optionValuePath="port"
                   optionLabelPath="port"
                   content=path.service.availablePorts
                   value=path.targetPort
                 }}
+              {{else if (eq path.backendType 'service')}}
+                {{new-select
```
vincent99: I'm not sure what the point of an empty select is here; if the path entry is valid without a targetPort then just don't have a select.  If it's not then disable the service entry so I can't select it in the first place?

PR #2418, opened 2018-11-19
Configmap

loganhz:
Proposed changes
======

- Fix the issue that download filename is always download.txt
- Support upload key value pairs when create config map
- Show import YAML button in config map list page as user can import it via YAML

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/16610
 

Further comments
======

### DIFF
--- a/app/authenticated/project/config-maps/index/template.hbs
+++ b/app/authenticated/project/config-maps/index/template.hbs
@@ -9,6 +9,9 @@
         {{#link-to (query-params group="namespace") classNames="btn btn-sm bg-default"}}<i class="icon icon-list-nested"></i>{{/link-to}}
       {{/tooltip-element}}
     </div>
+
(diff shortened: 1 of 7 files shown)


vincent99:
This seems pretty confusing, providing yaml in all other contexts means you're giving a k8s manifest file.

File types the input allows aren't restricted.

The mixin looks like copy-posted subset of input-text-file

loganhz:
@vincent99 

> This seems pretty confusing, providing yaml in all other contexts means you're giving a k8s manifest file.

Yes. User will be able to upload any k8s manifest file via `Import YAML` in Config Maps list page. 

But if we don't show the button here, users need to go to workloads list page and import config maps there. Does it make sense? 

image


> File types the input allows aren't restricted.

Changed it to `text/*`

image

> The mixin looks like copy-posted subset of input-text-file

Yes. Do you mean I should do `Component.extend(InputTextFile, {})` in `form-key-value`?

`InputTextFile` contains a lot of other attributes. Also it does something like `set(this, 'value', value);`
I guess it's better to have some mixin with simple logic for uploading ?

PR #2450, opened 2018-11-26
Fix a few french translations

such:
This fixes a few french translations (especially supprimer instead of supprimmer that was bugging me ;))


vincent99:
Only English and Chinese are maintained here; the "source of truth" for other languages is CrowdIn, https://translate.rancher.com (and then they are periodically synched to here).  Please submit these there.

PR #2462, opened 2018-12-05
Revert the fix for edit cluster page

loganhz:
Proposed changes
======
The correct fix is merged in https://github.com/rancher/ui/commit/0d41bb784cafea939f27d034432be83b6edef500

I need to remove my fix for it.

Types of changes
======

Linked Issues
======
 

Further comments
======


loganhz:
Well, it looks like this fix is correct. Don't need to revert it

PR #2480, opened 2018-12-07
Click on checkbox for bulk action doesn't work

n313893254:
Proposed changes
======

Types of changes
======

Linked Issues
======

https://github.com/rancher/rancher/issues/16601

 

Further comments
======


vincent99:
@westlywright this doesn't seem like it makes any sense, select all shouldn't friend on things being removable or bulk removable, that may not even be an action the resource has.

PR #2501, opened 2018-12-12
Support existing pvc as catalog question

loganhz:
Proposed changes
======
Support existing pvc as catalog question

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/16583
 

Further comments
======


loganhz:
There is no backend dependency as it is a new type for catalog question. We can merge it.

PR #2505, opened 2018-12-13
Support generate password by regex

loganhz:
Proposed changes
======
Support generate password by regex for catalog password answer

```
- group: "App Settings"
  variable: app.session.secret
  default: ""
  description: "secret will be auto-generated if not specified"
  min_length: 64
  max_length: 64
  valid_chars: "a-zA-Z0-9"
  type: password
  label: Session Secret
```

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/17055
 

Further comments
======

### DIFF
--- a/app/catalog-tab/launch/route.js
+++ b/app/catalog-tab/launch/route.js
@@ -124,7 +124,7 @@ export default Route.extend({
   },
 
   newNamespace(duplicateName, newNamespaceName) {
-    const suffix = randomStr(5, 'novowels');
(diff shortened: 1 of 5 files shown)


vincent99:
While it's not particularly well documented, that is not the [intended format](https://github.com/rancher/api-spec/blob/master/specification.md#character-ranges) for `validChars`, you'd just say `a-zA-z0-9` and probably shouldn't need an entire regex engine/library to generate them..

loganhz:
@vincent99 Updated. Thanks!

PR #2529, opened 2018-12-20
Support WeChat as notifier

loganhz:
Proposed changes
======
Now we don't support WeChat as a notifier. So user can not send the alert to wechat. We need to support it

Related PR
https://github.com/rancher/rancher/pull/17194

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/17121
 

Further comments
======


loganhz:
FYI @westlywright Backend PR has been merged.

PR #2532, opened 2018-12-21
Fix storage class default flag

vincent99:
Proposed changes
======
Fix setting the default flag on a storage class if the model doesn't already have an annotations object.

Types of changes
======

Bugfix

Linked Issues
======

rancher/rancher#15097


loganhz:
@vincent99 FYI the build is failed

PR #2534, opened 2018-12-21
Add current cpu/memory usage in cluster dashboard

n313893254:
Proposed changes
======

Cluster dashboard

image

Node dashboard

image

Types of changes
======

Linked Issues
======

https://github.com/rancher/rancher/issues/17040

 

Further comments
======


vincent99:
If live stats are enabled I think they should be the primary thing displayed.  So the large percentage would be 0.24/3 = 80%, etc.  And that would be main/blue part of the wheel as well.  And I'd call it "Used" instead of "Live".

Reserved would then be on the outside ring, and the ring should end before the end of the Used one since 5-10% of the total is not reservable (2.8 vs 3.0).

n313893254:
Updated

PR #2542, opened 2018-12-28
Fix form-key-to-path bug

n313893254:
Proposed changes
======

Types of changes
======

Linked Issues
======

https://github.com/rancher/rancher/issues/17199

 

Further comments
======

### DIFF
--- a/app/components/form-key-to-path/component.js
+++ b/app/components/form-key-to-path/component.js
@@ -113,10 +113,12 @@ export default Component.extend({
     if (secretName) {
       const secret = allSecrets.findBy('name', secretName);
 
-      set(this, 'keys', Object.keys(secret.data).map((k) => ({
-        label: k,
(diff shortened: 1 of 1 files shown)


app/components/form-key-to-path/component.js
```diff
@@ -113,10 +113,12 @@ export default Component.extend({
...
-      set(this, 'keys', Object.keys(secret.data).map((k) => ({
-        label: k,
-        value: k,
-      })));
+      if (secret) {
+        set(this, 'keys', Object.keys(secret.data).map((k) => ({
+          label: k,
+          value: k,
+        })));
+      }
```
vincent99: Keys doesn't get (re)set if the secret isn't found

n313893254:
@vincent99 Done, Thanks for your review.

PR #2546, opened 2019-01-03
Add advanced section for Service Discovery

loganhz:
Proposed changes
======

- Support for LB service creation with elaborate params
- Use `service` instead of `dnsRecord ` for Service Discovery tab. The fields of `dnsRecord` is a subset of `service`.
- The newly added fields are hidden by default such as Session affinity and Annotations.
- All of the extra fields in service are added in advanced section. User can config them via Rancher UI instead of YAML.

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/15070
 

Further comments
======
Today in Rancher UI there is only way to create an LB service for the workload - by setting up the port of LB type on the workload, so the underlying controller would create the LB service automatically.

The current limitation is you can't really set any extra parameters for the LB service like cloud provider specific annotations or session affinity. We should add support for it by allowing user to explicitly create a Load Balancer service with all the extra parameters via Services page, in addition to implicit-via-workload-ports way.


translations/en-us.yaml
```diff
@@ -910,6 +910,39 @@ editDns:
     view: "Record: {name}"
   name:
     placeholder: e.g. myapp
+  sessionAffinity:
+    title: Session Affinity
+    detail: Used to maintain session affinity
```
vincent99: `Map connections to a consistent target based on their source IP`

translations/en-us.yaml
```diff
@@ -910,6 +910,39 @@ editDns:
...
+    label: Enable client IP based session affinity
```
vincent99: Enable session affinity

loganhz:
@vincent99 Updated. Thanks for the review.

Only `An external hostname` is the special case as it is a service with kind `ExternalName`. The unrelated options are hidden now.
image

The other 4 options are all services with kind `ClusterIP`.  Now they are headless services. So the options which won't work for headless services should be hidden as well.
image

Users can only see the other options when select `Cluster IP`, `Node Port` or `Layer-4 Load Balancer`.

image
image
image

PR #2585, opened 2019-01-24
Use array for ingress rule path

loganhz:
Proposed changes
======
There is a breaking change in API. We have to use array for ingress rule path.
There is no UX change in UI.

Backend PR is https://github.com/rancher/rancher/pull/17129

We need to merge backend and frontend PRs together to make sure the feature is not broken in master build

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/17529
 

Further comments
======


cjellick:
@vincent99 gave me an offline LGTM. I'm going to go ahead and merge, since the backend is now merged itll break the UI

PR #2643, opened 2019-02-12
Provide option to set Administrative flag for cluster scoped roles

loganhz:
Proposed changes
======
Provide option to set Administrative flag for cluster scoped roles

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/17443
 

Further comments
======

### DIFF
--- a/lib/global-admin/addon/components/new-edit-role/template.hbs
+++ b/lib/global-admin/addon/components/new-edit-role/template.hbs
@@ -1,11 +1,11 @@
 <section class="header clearfix">
   <div class="pull-left">
     {{#if readOnlyBuiltInOrGlobal}}
-      <h1>{{t 'rolesPage.title'}}: {{model.role.name}}</h1>
+      <h1>{{t "rolesPage.title"}}: {{model.role.name}}</h1>
     {{else if editing}}
-      <h1>{{t 'rolesPage.editRole'}}</h1>
+      <h1>{{t "rolesPage.editRole"}}</h1>
     {{else}}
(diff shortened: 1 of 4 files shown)


vincent99:
I have no idea what the administrative flag is for from this.

loganhz:
@alena1108 @mrajashree what's the difference between `true` and `false` 
 for `Administrative` field?

loganhz:
From @mrajashree 

```
The name of the flag should be Administrative.
The reason behind adding this is, managing access to resources that are both, cluster scoped and project scoped, for example the `catalogtemplates` and `catalogtemplateversions`. If any cluster role has Administrative set to `true`, then that cluster role will get access to all catalogtemplates/versions from all projects that are a part of that cluster. Which is why cluster-owner has by default Administrative bit set to true; but cluster-member doesn't since cluster-members should not have access to catalogtemplates/versions from projects that they don't belong to, even if that project is in the cluster they are a part of. So in short this flag if set will grant cluster-scoped roles access to all resources across all projects in the clusters they are added to, and should be set only if the role being added is expected to be equivalent to a cluster-owner role
```


@vincent99 Any suggestion for the label and help text for that?

cjellick:
I explained this field to @vincent99 and we talked through the usecases and we decided that we dont want to expose this field in the UI. it should be set for our existing cluster admin roles, but we will just hide it. If people complain about custom roles, they can in the short term flip the bit in the API and then we can figure out the right way to explain what it does.

The problem is that it is fairly confusing and requires understanding how rancher does authz to fully grok

PR #2660, opened 2019-02-15
Consider missing cluster component status healthy, engine versions, dev proxy

vincent99:
Proposed changes
======

If a particular component status is missing from the cluster, call that component "healthy".  K3s and some cloud providers are missing some statuses.

Also support dev proxying to a L7 load balancer better

Types of changes
======

Enhancement


cjellick:
https://github.com/rancher/rancher/issues/18171

cjellick:
lgtm

cjellick:
for the docker version bit https://github.com/rancher/rancher/issues/18171

PR #2694, opened 2019-02-26
pinganyun ecs node-driver

ly5156:
Proposed changes
======
Add PingAnYun ECS Node Driver

Types of changes
======

Linked Issues
======
 

Further comments
======


vincent99:
Why is there nothing but translations here?  If this is a third party driver it should register its own translations (and region label key) in the driver.

loganhz:
There is no example for i18n in https://github.com/rancher/ui-driver-skel
Let me try if we can do that

vincent99:
this.get('intl').addTranslations('en-us', {key: value, ...});

PR #2758, opened 2019-03-13
Fix outer wheel greater than inner

n313893254:
Proposed changes
======

1. Fixed outer wheel greater than inner when reserved greater than used. 
2. Change styles according to https://github.com/rancher/rancher/issues/18553#issuecomment-469404081. 
3. I will change the circle to the triangle in the future.  @vincent99 

When reserved greater than used:
image

Normal situation:
image

Types of changes
======

Linked Issues
======

https://github.com/rancher/rancher/issues/18553
 

Further comments
======

### DIFF
--- a/lib/monitoring/addon/components/nodes-live/component.js
+++ b/lib/monitoring/addon/components/nodes-live/component.js
@@ -48,7 +48,11 @@ export default NodesReservation.extend(Metrics, {
       if (used === 0 || used === undefined) {
         set(this, 'cpuTotalPercent', 100)
       } else {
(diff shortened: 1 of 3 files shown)


vincent99:
This isn't the fix I was taking about, you're just capping it at 100%.  Both wheels should be based on the same scale, from 0 to max(2,4) = 4 for the CPU example.

Then the inner wheel gray should end at the top of the wheel (2 is half of the max of 4) because the rest is not possible (yet), and the green fill 54% of that gray area (which is the same as 1.1/4).

PR #2808, opened 2019-03-25
[2.3]Fix China hosted k8s issue

loganhz:
Proposed changes
======

- update ACK k8s version
- Fix Huawei cce edit issue

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/19148
https://github.com/rancher/rancher/issues/19147
 

Further comments
======


vincent99:
These are obviously going to continue to get out of date, can you just read them from the API instead of hardcoding in the UI?

loganhz:
@vincent99 Now we are using a sdk which only supports standard Alibaba Cloud API format.

However, the API for getting k8s version which is used in Alibaba Cloud console is not standard. And there is no document for it. So we didn't find out how to do the authentication stuff.

We have asked the Alibaba Cloud dev today if they have a standard API to get the versions. 

We will track it in https://github.com/rancher/rancher/issues/19152

PR #2809, opened 2019-03-25
use right env name for drone publish

drpebcak:


drpebcak:
@westlywright @cloudnautique

PR #2847, opened 2019-04-03
[2.2.2] Prometheus error when installing Monitoring

n313893254:
Proposed changes
======

As the backend API changed, the status is for watching Prometheus service. But no guarantee for node metrics is ready.
UI needs to Fallback to old dashboard view if monitoring is not healthy.

This PR can be merged any time. 

Types of changes
======

Linked Issues
======

https://github.com/rancher/rancher/issues/19326

 

Further comments
======


lib/monitoring/addon/components/nodes-live/component.js
```diff
@@ -22,8 +22,11 @@ export default NodesReservation.extend(Metrics, {
   nodes:             null,
   filters:           { resourceType: 'cluster' },
 
+  isGotMetrics: false,
   cluster:           reads('scope.currentCluster'),
```
loganhz: format it

lib/monitoring/addon/components/nodes-live/component.js
```diff
@@ -272,7 +296,7 @@ export default NodesReservation.extend(Metrics, {
     });
 
     const timeOutAnchor = setTimeout(() => {
-      this.send('query', false);
+      this.beforeQuery();
```
loganhz: the false is removed. Don't we need it?

PR #2884, opened 2019-04-14
[2.3]monitoring

loganhz:
Proposed changes
======
When `isMonitoringReady` status changed, UI will send requests to backend. In some setup, the conditions keep updating sometimes. But `MonitoringEnabled` condition didn't change at all. We should not send the requests in this case

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/19564
 

Further comments
======


loganhz:
backend in

PR #2930, opened 2019-04-26
Disable monitoring buttons if users don't have related permission

loganhz:
Proposed changes
======
Disable enable/save/disable button if users don't have related permission.

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/19074

 

Further comments
======


loganhz:
Updated

PR #2951, opened 2019-05-02
Add colon to templates that utilize generic.restarts and generic.podRestarts

davidnuzik:
Proposed changes
======

Add a colon to the end of "Pod Restarts" and "Restarts" (generic.podRestarts, generic.restarts) for improved formatting and readability. One template has this but they all should where needed.

Types of changes
======

Bugfix, non-breaking

Linked Issues
======
 
https://github.com/rancher/rancher/issues/17891

Further comments
======

Adds a colon in these areas:
Screen Shot 2019-04-30 at 2 34 34 PM
Screen Shot 2019-04-30 at 2 33 52 PM

Pod Restarts > Pod Restarts:
Restarts > Restarts:

### DIFF
--- a/app/components/accordion-container/template.hbs
+++ b/app/components/accordion-container/template.hbs
@@ -34,7 +34,7 @@
         <td data-title="{{dt.image}}">
           {{inst.image}}
         </td>
-        <td data-title="{{dt.restarts}}">
+        <td data-title="{{dt.restarts}}:">
           {{inst.restarts}}
         </td>
         <td data-title="{{dt.actions}}" class="actions">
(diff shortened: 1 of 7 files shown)


davidnuzik:
@loganhz I think this might actually break the container detail page. For some reason only there we manually add a colon it to the template. I should probably fix this just need to figure out where template is.

image

davidnuzik:
I think this is it https://github.com/rancher/ui/blob/master/app/container/template.hbs#L57

davidnuzik:
I got guidance from Wes @loganhz FYI I will modify the templates the use generic.restarts and generic.podRestarts so they now contain a colon.

loganhz:
Thanks! LGTM

PR #3001, opened 2019-05-20
Support one click upgrade option for cluster/project monitoring.

loganhz:
Proposed changes
======
Support one click upgrade option for cluster/project monitoring.

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/20304

 

Further comments
======

### DIFF
--- a/lib/monitoring/addon/cluster-setting/route.js
+++ b/lib/monitoring/addon/cluster-setting/route.js
@@ -1,11 +1,46 @@
 import Route from '@ember/routing/route';
-import { set } from '@ember/object';
+import { get, set } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { on } from '@ember/object/evented';
 import C from 'ui/utils/constants';
+import { hash } from 'rsvp';
 
 export default Route.extend({
   session:     service(),
+  scope:       service(),
+  globalStore: service(),
+
+  model() {
+    const store = get(this, 'globalStore');
+
+    const cluster = get(this, 'scope.currentCluster');
+    const project = get(cluster, 'systemProject');
+
+    if ( !project || !get(cluster, 'enableClusterMonitoring') ) {
+      return { apps: [], }
+    }
+
+    return hash({
+      apps: store.rawRequest({
+        url:    `/v3/project/${ get(project, 'id') }/apps`,
+        method: 'GET',
+      }).then((res) => {
+        const out = [];
+        const apps = get(res, 'body.data') || [];
+        const clusterApp = apps.findBy('name', 'cluster-monitoring');
+        const operatorApp = apps.findBy('name', 'monitoring-operator');
+
+        if ( clusterApp ) {
+          out.push(store.createRecord(clusterApp));
+        }
+        if ( operatorApp ) {
+          out.push(store.createRecord(operatorApp));
+        }
+
+        return out;
+      }),
+    });
+  },
 
   setDefaultRoute: on('activate', function() {
     set(this, `session.${ C.SESSION.CLUSTER_ROUTE }`, 'authenticated.cluster.monitoring.cluster-setting');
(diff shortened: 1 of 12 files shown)


lib/monitoring/addon/cluster-setting/route.js
```diff
@@ -1,11 +1,46 @@
...
+    const cluster = get(this, 'scope.currentCluster');
+    const project = get(cluster, 'systemProject');
+
+    if ( !project || !get(cluster, 'enableClusterMonitoring') ) {
+      return { apps: [], }
+    }
+
+    return hash({
+      apps: store.rawRequest({
+        url:    `/v3/project/${ get(project, 'id') }/apps`,
```
westlywright: Is there a reason this can not be fetched using the project store?

lib/shared/addon/mixins/catalog-upgrade.js
```diff
@@ -0,0 +1,69 @@
+import Mixin from '@ember/object/mixin';
```
westlywright: There is a mixin `upgrade-component` that could possibly be used here. It would take some modification for your action but it does already have status and parsed versions and version links. If it doesn't make sense to use that rather then create a new mixin, is there a reason we're re-exporting it out of the engine in the corresponding app file? It doesn't look like its used outside of the monitoring engine so if it is not I'd like to remove the re-export so we're not bulking up our dist with unnecessary plumbing required for re-export.

lib/shared/app/mixins/catalog-upgrade.js
```diff
@@ -0,0 +1 @@
+export { default } from 'shared/mixins/catalog-upgrade';
```
westlywright: See comment in main component file.

lib/monitoring/addon/cluster-setting/route.js
loganhz: It's in cluster level menu, so no project is selected yet. I need to use the project store of System project in the cluster.

It looks like we can change the baseUrl if we want to use project store. Does it make sense?

```
    const projectStore = get(this, 'store')

    set(projectStore, 'baseUrl', `${ get(this, 'app.apiEndpoint') }/projects/${ project.id }`);
    const apps = projectStore.find('app', null, { forceReload: true })
```

loganhz:
Updated.  @westlywright 

Now it is using the `latestVersion` and `upgradeStatus` fields from `upgrade-component`.
`catalog-upgrade` has been moved into monitoring lib.
Using projectScope instead of sending raw requests to fetch apps.

lib/monitoring/addon/cluster-setting/route.js
westlywright: I'm sorry you changed it but with your explanation the way you disigned it originally makes more sense. It was not clear why that needed to be done that way though. Perhaps a small inline comment for any maintainers as to why we are using that would be helpful? I don't think changing the `baseUrl` is the right move though. That would change it for everything in the project store, which when you select a project should get fixed but we potentially introduce more issues. Id rather go back to rawRequest because now it makes sense.

loganhz: @westlywright No worries! But if we use rawRequest, I have another problem.

`get(app, 'externalIdInfo'); ` will return undefined as `externalIdInfo` field is in `models/app.js`.

How can I convert the json object to a model if I use rawRequest?

loganhz:
image

@davidnuzik does this message make sense?

davidnuzik:
@loganhz Yes it makes sense to me. If I were picky though, I might suggest removing the period at the end.

davidnuzik:
"Upgrade monitoring to the latest version"
Is better

loganhz:
@westlywright updated. Now it is using rawRequest.

lib/monitoring/addon/cluster-setting/route.js
westlywright: Could you use the `_add(type, obj)` method on the store?

loganhz: Thanks! All I need is `createRecord`

PR #3016, opened 2019-05-24
update pt-br translation

paulobezerr:
Proposed changes
======

I have tried to use `scripts/update-translations` script but did not work.
Maybe this is why I did not see translation result in new UI version.

Then, I decided to send a pull request with updated PT-BR translations.

Types of changes
======

- Translations

### DIFF
--- a/translations/pt-br.yaml
+++ b/translations/pt-br.yaml
@@ -10,6 +10,7 @@ generic:
   any: Qualquer
   application: aplicação
   as: como
+  back: Voltar
   cancel: Cancelar
   closeModal: Fechar
   cluster: Cluster
@@ -33,6 +34,7 @@ generic:
   done: Concluído
   download: Baixar
   edit: Editar
+  emptyString: ""
   enable: Habilitar
   enabled: Ativado
   entrypoint: Ponto de entrada
@@ -44,6 +46,8 @@ generic:
   hardware: Hardware
   healthy: Sem erros
   help: Ajuda
+  hostname: Nome do servidor
+  hours: horas
   id: Identificador
   image: Imagem
   import: Importar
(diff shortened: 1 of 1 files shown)


vincent99:
Only English and Chinese are maintained by us in-repo.  The rest are maintained in https://crowdin.com/project/rancher-ui and periodically pulled into here (which is what `update-translations` does).

paulobezerr:
Well, there is a long time without updates in pt-br translations.
Can we automate this?

vincent99:
They are updated usually before major releases.  What's sitting there and not approved is a tiny fraction of the missing translations for pt-br, so it's not exactly going to make a huge dent even if it were merged.

paulobezerr:
@vincent99, honestly speaking, keep working on translations and see no changes but in major releases demotivates.

This pull request is huge and only have Crowdin translations.
I know that is "only" 26% of translations, but it is already lot.

IMHO this should be updated in every version for all languages.

vincent99:
There are **16** strings in CrowdIn pending approval that are not already in master.  I don't know how long they've been pending off-hand, but out of ~5000 that's basically noise.

This PR obviously contains more, but CrowdIn is the mechanism by which we manage translations for all languages other than en-us and zh-hans, not PRs.  It also contains thousands of lines of duplicate English keys.  English is already the fallback if there is not a translation found in the selected language, it doesn't need to go in the translation files N times.  The translation file should include only the keys that are actually translated.

paulobezerr:
Got it, sir.

But maybe we should check what is happening, because translations accepted 2 mounts ago are not in master.

### Example

The term `Saving...` was translated to PT-BR as `Sanvando...`.
But two months ago I have accepted the correct one: `Salvando...`.

image

This term isn't in master yet.

Could you help on it?
Can we accept translations in a shorter time?

PR #3028, opened 2019-06-04
Istio

loganhz:
Proposed changes
======
* Fix the issue that switch project doesn't update the metrics
* Don't allow to visit istio page if cluster is not ready

Types of changes
======

Linked Issues
======
rancher/rancher#20626
 

Further comments
======


loganhz:
Updated. Changed the solution. Now istio menu is disabled if cluster is not ready.
So we don't need to add the redirect logic now.

I think it deserves another issue if we would like to refine the style for disabled menu.

PR #3049, opened 2019-06-12
[Preview 2 Release]Show warning for registry admin password

loganhz:
Proposed changes
======
Now if users re-use the PVC, the admin password will not take effect. 

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/20797

 

Further comments
======

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -1209,6 +1209,7 @@ globalRegistryPage:
     adminPassword:
       label: Admin Password
       placeholder: Harbor Admin password
+      helpText: "If you are using an existing PVC with initialized Harbor data, the password will not take effect."
   config:
     header:
       postgresql: Database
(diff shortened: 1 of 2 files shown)


translations/en-us.yaml
```diff
@@ -1209,6 +1209,7 @@ globalRegistryPage:
     adminPassword:
       label: Admin Password
       placeholder: Harbor Admin password
+      helpText: "If you re-use an existing persistent volume claim for Database component, the admin password will not take effect."
```
gitlawr: The password will take effect, if the existing PVC bounds to a fresh PV.
See case 2 in the comment: https://github.com/rancher/rancher/issues/20797#issuecomment-501515382

loganhz: How about `If you are using the existing PVC which bounds to a used PV, the password will not take effect`

gitlawr: I prefer `If you are using the existing PVC with initialized Harbor data, the password will not take effect`

loganhz:
Updated

gitlawr:
LGTM

PR #3063, opened 2019-06-17
fix istio versioning

deniseschannon:
Proposed changes
======

Update UI for istio to match the version that the catalog will be 

Types of changes
======

Linked Issues
======
 

Further comments
======


cjellick:
LGTM

PR #3149, opened 2019-07-22
Add Artifactory as an option for registry

loganhz:
Proposed changes
======
Show Artifactory as an option for registry
There is no difference with custom.
nothing except that list aritfactory as a radio button

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/20394

 

Further comments
======

### DIFF
--- a/app/components/cru-registry/template.hbs
+++ b/app/components/cru-registry/template.hbs
@@ -71,6 +71,21 @@
               {{t "cruRegistry.address.quay"}}
             </label>
           </div>
+          <div>
+            <label>
+              {{radio-button
+                selection=row.preset
+                value="artifactory"
+              }}&nbsp;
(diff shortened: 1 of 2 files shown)


loganhz:
> Does this need a new icon? How does the UI handle displaying a Artifactory entry without it?

No, we don't need. Actually we only can see the Artifactory option when creating a registry credential. If we want to display it in registry credentials list, we have to add a new field in backend.

Not sure if it is good enough. @deniseschannon

deniseschannon:
We should always be able to see the artifactory option. As it can either be deployed through a catalog app or it could be hosted on artifactory side. 

@loganhz

loganhz:
Yes, we will always be able to see the artifactory option in adding registry page. @westlywright I think there is nothing we need to change in this PR

PR #3211, opened 2019-08-08
cityscape

lvuch:
change login to match baunfire branding


vincent99:
@lvuch There's no file add for `login-cityscape.svg` in here?  And delete the old `login-container-farm.svg` file unless it's still used somewhere.

lvuch:
it was added in this pr: https://github.com/rancher/ui/pull/3204/files

PR #3302, opened 2019-08-27
[2.3.0]Use disable/enable actions on ClusterTemplateRevision object

n313893254:
Proposed changes
======

Use disable/enable actions on ClusterTemplateRevision object instead of  using `save()`.

Types of changes
======

Linked Issues
======

https://github.com/rancher/rancher/issues/22374

 

Further comments
======


n313893254:
@vincent99 Done, thanks for your review.

PR #3308, opened 2019-08-28
18502

lvuch:
https://github.com/rancher/rancher/issues/18502

### DIFF
--- a/app/styles/components/_forms.scss
+++ b/app/styles/components/_forms.scss
@@ -94,14 +94,14 @@ textarea {
     }
 
     &::placeholder {
-        color: $input-color-placeholder;
+        color: rgba($input-color-placeholder, .35);
     }
 
     &:hover {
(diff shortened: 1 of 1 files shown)


vincent99:
@lvuch The issue is asking for the regular text color to be white(r) (not to change the placeholder, though you can update those too)

davidnuzik:
No, we are actually expecting this:

image

This is how it used to be a while back; then something changed. The issue describes this problem. This is the field at rest with no input (and not selected) then different text color of placeholder text when click in the box. Then different text for typed in text by the user.

lvuch:
@davidnuzik 
Screen Shot 2019-08-29 at 12 15 37 PM
Screen Shot 2019-08-29 at 12 15 29 PM
Screen Shot 2019-08-29 at 12 15 24 PM
Screen Shot 2019-08-29 at 12 15 17 PM

PR #3320, opened 2019-08-30
update zh-hans

xiaoluhong:
Proposed changes
======
Update zh-hans 

Types of changes
======

Linked Issues
======
 

Further comments
======

In the final `services` configuration, I added some K8S cluster configuration parameters. Some of these parameters are commonly used by users and some are configured according to best practices. Many users don't know how to configure these parameters, and our UI doesn't currently support customizing these parameters, so I put these common parameters into translation so that users can easily know how to use the parameters and how to configure them as they edit cluster yaml.


vincent99:
@xiaoluhong Conflicts with https://github.com/rancher/ui/pull/3317/files ; I'd just resolve it but have no idea which is better...

xiaoluhong:
Hi @vincent99 ,  I have resolved the conflict.  thank you.

PR #3334, opened 2019-09-06
Strip off https:// for private registry

loganhz:
Proposed changes
======
Strip off https:// for private registry

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/21591

 

Further comments
======

### DIFF
--- a/app/styles/components/_forms.scss
+++ b/app/styles/components/_forms.scss
@@ -66,6 +66,11 @@ textarea:required {
   box-shadow: none !important;
 }
 
+// Firefox fix to turn off the red box shadow on invalid fields
+input:invalid {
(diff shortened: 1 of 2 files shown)


app/styles/base/_helpers.scss
```diff
@@ -326,3 +326,7 @@ a.disabled {
 .pointer-events-none {
   pointer-events: none;
 }
+
+.no-invalid-box:invalid {
```
vincent99: This should go in _forms.scss and apply to all inputs instead of this particular one. There are probably others and firefox is inconsistent with every other browser.

loganhz:
Updated. Thanks!

PR #3340, opened 2019-09-10
Support search in apps and multi-cluster apps list page

loganhz:
Proposed changes
======
Support search in apps and multi-cluster apps list page

Types of changes
======

Linked Issues
======
https://github.com/rancher/rancher/issues/22744
 

Further comments
======

### DIFF
--- a/lib/shared/addon/components/search-text/component.js
+++ b/lib/shared/addon/components/search-text/component.js
@@ -1,0 +1,14 @@
+import Component from '@ember/component';
+import { set } from '@ember/object';
+import layout from './template';
+
+export default Component.extend({
+  layout,
+  searchFields: ['displayName', 'id:prefix', 'displayState'],
+
+  actions: {
+    clearSearch() {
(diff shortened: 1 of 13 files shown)


lib/shared/addon/components/search-text/component.js
```diff
@@ -0,0 +1,46 @@
+import Component from '@ember/component';
+import { set } from '@ember/object';
+import layout from './template';
+import { matches } from 'shared/components/sortable-table/component';
+
+const SEARCH_FIELDS = ['displayName', 'id:prefix', 'displayState'];
+
+export function filter(array, searchText, searchFields = SEARCH_FIELDS) {
```
vincent99: This looks to be copied from sortable-table (and apparently `pod-dots` and `namespace-table` do the same thing)?  I'd like to see is pulled out as its own utility and then all components use that instead of adding a 4th copy-paste of the logic to filter an array by search expression.

loganhz:
Updated. Thanks!

PR #3351, opened 2019-09-12
Add standard volume type for AWS

MeikelLP:
Proposed changes
======

Add [`standard`](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html) as storage type for AWS EC2.

Types of changes
======

Added storage type `standard` to the list of available storage types.

Further comments
======

In our company we require storages to be type of `standard` which was not possible with rancher so far.

We'd also like to add this to the storage classes page but it's not possible due to a [bug/missing feature in Kubernetes](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/365) (will throw errors/warnings if you try). The first commit contains this but was reverted in commit 2.

### DIFF
--- a/lib/nodes/addon/components/driver-amazonec2/component.js
+++ b/lib/nodes/addon/components/driver-amazonec2/component.js
@@ -8,11 +8,12 @@ import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import NodeDriver from 'shared/mixins/node-driver';
 import layout from './template';
-import { INSTANCE_TYPES, nameFromResource, tagsFromResource, REGIONS } from 'shared/utils/amazon';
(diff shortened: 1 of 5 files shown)


vincent99:
rancher/rancher#22898

vincent99:
@MeikelLP I'm reverting this because it's presenting invalid options for the situation (can't do cold storage for a boot disk, and provisioned IOPS probably needs to specify how many) and we're too near the end of the 2.3 cycle. See https://github.com/rancher/rancher/issues/22898.

MeikelLP:
Is this an issue I need to fix in UI or is this an internal rancher thing?

PR #3352, opened 2019-09-12
Add more context to the title of deletion modals

codyrancher:
Adds the resource types "namespace", "cluster", "project" as
    additional context to the titles of the respective deletion modals
    in order to reduce the likelyhood of a user accidentally deleting
    an unintented resource.
    
    We changed the title from "Are you sure you want to delete" to
    "Are you sure you want to delete the namespace:" with namespace
    being bold. The same is true for cluster and project as well.
    
    rancher/rancher#17619

The change where the deletion occurs on a namespace resource type:
Screen Shot 2019-09-12 at 11 23 56 AM

The change where the deletion occurs on a project resource type:
Screen Shot 2019-09-13 at 12 34 14 PM

The change where the deletion occurs on a cluster resource type:
Screen Shot 2019-09-13 at 12 34 50 PM

The change where the deletion occurs on a non-elevated resource type:
Screen Shot 2019-09-13 at 12 37 31 PM

### DIFF
--- a/lib/shared/addon/components/confirm-delete/component.js
+++ b/lib/shared/addon/components/confirm-delete/component.js
@@ -8,6 +8,12 @@ import layout from './template';
 import { eachLimit } from 'async';
 import C from 'ui/utils/constants';
 
+function hasSomeOfResourceType(resourceType) {
+  console.log(get(this, 'resources'));
+
+  return get(this, 'resources')
+    .some((resource) => get(resource, 'type') === resourceType);
+}
 
 export default Component.extend(ModalBase, {
   settings:       service(),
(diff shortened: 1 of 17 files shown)


codyrancher:
> * This wasn't really well described in the issue, but seems like the bold message should be shown for deleting namespaces or "higher" (project, cluster).  And show for all namespaces, not just the system one.  (System gets the extra info banner, which still makes sense).
1. I've removed the system constraint for namespace modals.
2. I've added project and cluster to this change. I failed to find an enumeration of resource types to judge for myself but if there are any additional resource types I should add feel free to let me know.
> * Don't use "fixes #x" / "resolves #x" / etc; Issues get moved to QA once merged (the `[zube]: To Test` label) and then they close it once validated.  Linking to the issue in the PR is good though, the person merging it will go through the issue links and move their label.
1. I've removed resolves and left the issue link.
> * We don't normally do the translations (other than `en-us`).  `zh-*` are maintained by our team in China as their own PRs, for all the others https://translate.rancher.com is the source of truth and they are synced into github
1. I didn't provide any translations beyond en-us. I did change a string id which I carried through.

vincent99:
Oops I missed that only the key changed on the translations.. 👍

PR #3382, opened 2019-09-24
Include Template/Revision Upgrade Notifications

codyrancher:
Proposed changes
======
We want to make it easier for someone that is managing multiple
clusters to see what template and template revision a cluster
is using. We also want to make it easy for a user to notice
that a cluster isn't using the latest revision.

To achieve this we added the template and revision to a column
in the clusters table to display both. We also indicate add
a warning badge beside a revision when the revision isn't
the latest.

Screenshots:
Screen Shot 2019-09-24 at 12 24 33 PM
Screen Shot 2019-09-24 at 12 24 20 PM

Types of changes
======

Linked Issues
======
rancher/rancher#22047


codyrancher:
Please hold up on merging this until I get some feedback on the original issue.

PR #3449, opened 2019-10-15
Improve vsphere driver UX

codyrancher:
Proposed changes
======
-We now prepopulate fields as select component instead of just
free-form text fields.
-We added support for cloning from different sources.
-We also added support for custom attributes and tags.

To implement the above features we also added support for promise
based content for new-select and search-select and created a new
component contextual-select-array which provides a method for
filtering content based on previous selections. We also added content
support for the key-value-array component.

Types of changes
======

Linked Issues
======
rancher/rancher#21580
rancher/rancher#23031
rancher/rancher#23002
rancher/rancher#19032

### DIFF
--- a/.ember-cli
+++ b/.ember-cli
@@ -9,6 +9,7 @@
   "port"             : 8000,
   "host"             : "0.0.0.0",
   "usePods"          : true,
+  "liveReload"       : true,
   "liveReloadPort"   : 37531,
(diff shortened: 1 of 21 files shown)


codyrancher:
Closing in favor of https://github.com/rancher/ui/pull/3475

PR #3521, opened 2019-10-31
Switch node sortkey on Workload page.

codyrancher:
Proposed changes
======
The node sortkey was displayname instead of node.displayName which
caused sorting to behave strangely.

I switched the correct sortkey of node.displayName.

Types of changes
======

Linked Issues
======
rancher/rancher#21218

Screen Shot 2019-10-30 at 5 23 56 PM

### DIFF
--- a/app/components/accordion-pod/component.js
+++ b/app/components/accordion-pod/component.js
@@ -13,24 +13,24 @@ export default Component.extend(ManageLabels, {
   headers:       [
     {
       name:           'displayState',
-      sort:           ['displayState'],
+      sort:           ['displayState', 'sortName', 'id'],
       translationKey: 'generic.state',
       width:          120
     },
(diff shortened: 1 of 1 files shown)


vincent99:
It should be sorted by `[node.]sortName`, which [looks for](https://github.com/rancher/ui/blob/master/lib/shared/addon/mixins/cattle-transitioning-resource.js#L454-L456) numeric suffixes in `displayName` and sorts them naturally instead of 1 10 11 2 3 4).

vincent99:
There should also be a 2nd column to sort on (for all the headers in that file...) so that there is always something unique in the sorted columns.  Usually the `id` of the thing the row is for as a last resort.

As it currently is the sort for all the pods that have the same node is unstable and effectively random.

PR #3568, opened 2019-11-11
[backport] Enable '.' in answer variables

codyrancher:
Proposed changes
======
This enables '.' to be a valid character in answer varaibles.
So we can support sub.key as a key now.

This also resolves a double render issue that was ran into while
reproducing this issue.

Types of changes
======

Linked Issues
======
rancher/rancher#24025

Further comments
======
Screen Shot 2019-10-17 at 12 09 47 PM


deniseschannon:
https://github.com/rancher/rancher/issues/23859

PR #3627, opened 2019-12-12
Ensure awsCloudProvider presense for custom cluster

codyrancher:
Proposed changes
======
awsCloudProvider was being removed from custom clusters when editing
the YAML. This change ensures that the field is still present. It also
required modifying the removeEmpty method as it was removing fields that
were in the excludedKeys argument.

Types of changes
======

Linked Issues
======
rancher/rancher#24515

### DIFF
--- a/lib/shared/addon/utils/util.js
+++ b/lib/shared/addon/utils/util.js
@@ -436,8 +436,8 @@ function keyNotType(k) {
 
 export function removeEmpty(obj, excludedKeys = []){
   return Object.keys(obj)
-    .filter((k) => !isEmpty(obj[k]) && (typeof obj[k] !== 'object' || keyNotType(obj[k])))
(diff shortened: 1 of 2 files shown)


lib/shared/addon/utils/util.js
```diff
@@ -436,7 +436,7 @@ function keyNotType(k) {
 
 export function removeEmpty(obj, excludedKeys = []){
   return Object.keys(obj)
-    .filter((k) => !isEmpty(obj[k]) && (typeof obj[k] !== 'object' || keyNotType(obj[k])))
+    .filter((k) => excludedKeys.indexOf(k) !== -1 || (!isEmpty(obj[k]) && (typeof obj[k] !== 'object' || keyNotType(obj[k]))))
```
vincent99: There are still places using indexOf / -1 but most checks have been changed to `excludedKeys.includes(k)` because it's easier to read and not get wrong.

codyrancher: Fixed and updated the other existing indexOf reference in the method.

PR #3642, opened 2019-12-17
Make special case for the 'running' state color

codyrancher:
Proposed changes
======
I originally changed the color in transitioning which changed the
color elsewhere like the workload progress bars. We decided to
make the color of running an exceptional case in CIS scans since the
meaning of running differs here.

Types of changes
======

Linked Issues
======
rancher/rancher#24440
rancher/rancher#24639

### DIFF
--- a/app/models/clusterscan.js
+++ b/app/models/clusterscan.js
@@ -4,7 +4,7 @@ import moment from 'moment';
 import { downloadFile } from 'shared/utils/download-files';
 import ObjectsToCsv from 'objects-to-csv';
 
-export default Resource.extend({
+const ClusterScan = Resource.extend({
   type:          'clusterScan',
   report:        'null',
   reportPromise: null,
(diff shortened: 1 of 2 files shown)


vincent99:
Why are we painting over this to ignore transitioning instead of fixing the API?  It's correctly red because transitioning=error, the data is wrong.

codyrancher:
It's no longer an issue with transitioning=error. 

The color for cattle-transitioning-resource is text-success which makes the running state green.
The desire is for the running state to be text-info or blue in CIS per the original design. 

The only other option I see is to change the state from running to something else or to create a new state like "Scanning".

vincent99:
Those are defaults, there's already support for per-type state mapping.

```javascript
// app/models/clusterscan.js

ClusterScan.reopenClass({
  stateMap: {
    'running': {color: 'text-info', icon: 'something'}
  }
});
```

PR #3650, opened 2019-12-19
Conditionally show node links in CIS Scans

codyrancher:
Proposed changes
======
If nodes have been deleted previous reports with missing nodes were
breaking. To resolve that we now only show links to nodes that we can
find.

Types of changes
======

Linked Issues
======
rancher/rancher#24676

Screen Shot 2019-12-19 at 12 12 51 PM


codyrancher:
@sowmyav27 is checking to see if the changes to links is acceptable before we merge this.

PR #3720, opened 2020-01-19
Add Switzerland region to Azure cluster provider

hypnotortoise:
Proposed changes
======
Rancher UI support for selecting Azure Switzerland regions. Because this region is not available without switching to paying account & enable as quota, added this as a new cloud region same as  China/Germany/US-Gov.

Types of changes
======
New feature

Linked Issues
======
 https://github.com/rancher/rancher/issues/25019

Further comments
======


mscbpi:
Is it normal not to see Switzerland in Rancher 2.4.8 ?

Thanks

bashofmann:
The swiss regions should be available starting with Rancher 2.5.0.

PR #3931, opened 2020-04-22
Add Oracle Container Engine (OKE) as a builtin custom cluster driver.

jlamillan:
Proposed changes
======

The PR adds Oracle Container Engine (OKE) [cluster driver](https://github.com/rancher-plugins/kontainer-engine-driver-oke) as a builtin custom cluster driver that will live in the UI repository. Once this is merged, the plan is to archive the [previous location](https://github.com/rancher-plugins/ui-cluster-driver-oke).

cluster

Types of changes
======

Improves usability of OKE driver in a backwards compatible manner. 

Linked Issues
======

- https://github.com/rancher/rancher/issues/20132
- [PR](https://github.com/rancher/rancher/pull/26664) to make OKE cluster driver builtin
 

Further comments
======
Moving the UI into this repo is a followup from the review [here](https://github.com/rancher/rancher/pull/26664). It also makes this [previous PR](https://github.com/rancher/ui/pull/3918) obsolete.


lib/shared/addon/components/cluster-driver/driver-oracleoke/component.js
```diff
@@ -0,0 +1,448 @@
...
+  'Toronto':   'ca-toronto-1',
+  'Frankfurt': 'eu-frankfurt-1',
+  'Zurich':    'eu-zurich-1',
+  'Sao Paolo': 'sa-saopaulo-1',
+  'London':    'uk-london-1',
+  'Ashburn':   'us-ashburn-1',
+  'Phoenix':   'us-phoenix-1',
+}
+
+const k8sVersionMap = {
```
vincent99, line 24: Is any of this const info available from your API?  We typically pull this kind of stuff from there instead of hardcoding.

You're going to be sort of painting yourselves into a corner when ask this is burnt into versions that users don't necessarily update frequently and you add a new region or don't want to support k8s 1.14.x anymore.

lib/shared/addon/components/cluster-driver/driver-oracleoke/component.js
```diff
@@ -0,0 +1,448 @@
...
+  'VM.GPU3.8':               'VM.GPU3.8',
+}
+
+const imageMap = {
+  'Oracle-Linux-7.6': 'Oracle-Linux-7.6',
+  'Oracle-Linux-7.5': 'Oracle-Linux-7.5',
+  'Oracle-Linux-7.4': 'Oracle-Linux-7.4',
+}
+
+const languages = {
```
vincent99, line 69: This can just go in the regular `translations/ en-us` file now instead of dynamically loading in here.

lib/shared/addon/components/cluster-driver/driver-oracleoke/component.js
```diff
@@ -0,0 +1,448 @@
...
+  }),
+  k8sVersionChoices: Object.entries(k8sVersionMap).map((e) => ({
+    label: e[1],
+    value: e[0]
+  })),
+  k8sUpgradeVersionChoices: computed('config.kubernetesVersion', function() {
+    let supportedVersions = Object.assign({}, k8sVersionMap);
+    var currentVersion = get(this, 'config.kubernetesVersion');
+
+    currentVersion.split('.').map((v) => {
```
vincent99, line 314: This does nothing because the map return value isn't used?

lib/shared/addon/components/cluster-driver/driver-oracleoke/component.js
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    currVer = currVer.substr(1);
+    toVer = toVer.substr(1);
+    currVer = currVer.split('.');
+    toVer = toVer.split('.');
+
+    var len = Math.max(currVer.length, toVer.length);
+
+    for (var i = 0; i < len; i++) {
```
vincent99, line 438: It doesn't necessarily matter for any current versions but these will not work right because the parts are still strings and e.g. `"2" > "10"`. 

Also see https://github.com/rancher/ui/blob/master/lib/shared/addon/utils/parse-version.js#L102

lib/shared/addon/components/cluster-driver/driver-oracleoke/component.js
jlamillan: I totally agree that we should be pulling in these values dynamically. At some point, I decided to submit that changes as a _subsequent enhancement_ to the driver(s). But your point about users not updating frequently is well taken.

_My preference_ would be to make the enactment as a part this PR. On the other hand, it is arguably an enhancement and there is a desire to have both of the drivers (OKE cluster, and OCI node) on the next release train....

jlamillan: We agreed offline that pulling in these values dynamically is desirable, but technically not required and therefore could be submitted as a subsequent _enhancement_.

PR #3994, opened 2020-05-21
[backport 2.3] Fix registry credentials so that it doesn't crash 

loganhz:
Proposed changes
======
Object destructuring expects the entire object to be present. When
'Available to all namespaces in this project' is used the namespace
object isn't defined which causes the save to fail due to an null
exception.

Types of changes
======

Linked Issues
======
rancher/rancher#27182
 

Further comments
======


maggieliu:
related to: https://github.com/rancher/rancher/issues/27182

PR #3997, opened 2020-05-21
Splitting up aggressive drain parameters   

codyrancher:
We no longer have safe and aggressive nodes we now
have two properties 'force' and 'deleteLocalData'.

Types of changes
======

Linked Issues
======
rancher/rancher#27111

### DIFF
--- a/app/components/modal-drain-node/component.js
+++ b/app/components/modal-drain-node/component.js
@@ -10,7 +10,7 @@ export default Component.extend(ModalBase, {
   growl: service(),
 
   layout,
-  classNames: ['large-modal'],
+  classNames: ['medium-modal'],
 
   selection: {},
 
--- a/lib/shared/addon/components/cluster-driver/driver-rke/component.js
+++ b/lib/shared/addon/components/cluster-driver/driver-rke/component.js
@@ -970,14 +970,12 @@ export default InputTextFile.extend(ManageLabels, ClusterDriver, {
     const maxUnavailableControlplane = get(source, 'maxUnavailableControlplane');
     const maxUnavailableWorker = get(source, 'maxUnavailableWorker') || '';
     const maxUnavailableUnit = maxUnavailableWorker.includes('%') ? 'percentage' : 'count';
-    const deleteLocalData = get(source, 'nodeDrainInput.deleteLocalData');
     const drainInput = get(source, 'nodeDrainInput');
 
     set(this, 'upgradeStrategy.maxUnavailableControlplane', maxUnavailableControlplane);
     set(this, 'upgradeStrategy.maxUnavailableWorker', maxUnavailableWorker.replace('%', ''));
     set(this, 'upgradeStrategy.maxUnavailableUnit', maxUnavailableUnit);
     set(this, 'upgradeStrategy.nodeDrainInput', drainInput ? drainInput : {});
-    set(this, 'upgradeStrategy.nodeDrainInput.deleteLocalData', (typeof deleteLocalData !== 'undefined') ? deleteLocalData.toString() : deleteLocalData);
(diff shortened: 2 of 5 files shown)


codyrancher:
@westlywright you were that both parameters should be there so I went ahead and made that change.

I spoke with @sowmyav27 and we don't have any documentation pointing out what these values should be. However, the backend hasn't changed the way these are handled so it should've remained the same from 2.3 to 2.4 so it's safe to assume 2.3 is correct.

codyrancher:
I'm adding @vincent99 since Wes is out right now.

vincent99:
The way it worked [before](https://github.com/rancher/ui/commit/68577cfcaa2432d3082edee848509a6ad518533f#diff-ed35d4c58d595a310737947665bac2ed) just had properties the user manipulated and  produced a set of options once at the end when you clicked to execute the action (and also used to set `force`...), because it was a one-shot modal.

I get that this was refactored to reuse in RKE; but there's now persisted state backing that, which the user might have specifically chosen.  By combining force + deleteLocalData into one aggressive flag you're silently overwriting their choice of `deleteLocalData` to be equal to `force`, via `aggressive`.  

That's maybe(?) ok in this case, but should be at least noted because we're talking about permanently deleting data here.  There's also the templating problem of the user could have permission to edit only one of those fields.  So you may just need separate direct controls for these flags in the RKE config case (or conditionally, or refuse to edit as form if in state you don't like).

--- 

The whole variables/naming/call sequence of basically

```javascript
var aggressive = false;
aggressive = force;
// updateAggressive fires. maybe
force = aggressive;
deleteLocalData = aggressive
// but both of those are exactly the same as updateAggressive

// user manipulates aggressive via form
// updateAggressive watches aggressive, updates force and deleteLocalData
// mode watches force and updates displayed label (to the aggressive translation)
```

seems unnecessarily complicated and circuitous; `force` is meant to equal `aggressive` at all times, so why not just have everything watch and edit it?

codyrancher:
> That's maybe(?) ok in this case, but should be at least noted because we're talking about permanently deleting data here. 

Would you like it noted somewhere other than the code?

> There's also the templating problem of the user could have permission to edit only one of those fields.  So you may just need separate direct controls for these flags in the RKE config case (or conditionally, or refuse to edit as form if in state you don't like).

The only difference between the old and the new is when that data is being manipulated. I don't think that difference is significant here because data will only change when something is saved or drained. Maybe you mean that someone could inadvertently change the value when editing a cluster and making a mode selection? I can agree with that but to fix this I think we should change the UI to expose the props independently. 

> seems unnecessarily complicated and circuitous; force is meant to equal aggressive at all times, so why not just have everything watch and edit it?

I prefer explicitly associating the values with aggressive so it doesn't later unintentionally become disassociated with that concept. The only reason it's circular is because the value can be undefined. I could remove that circular part though.

vincent99:
> I can agree with that but to fix this I think we should change the UI to expose the props independently.

Yes, that's what I meant by separate direct controls.  I think that's the safest thing to do for the RKE config; a support call where we unexpectedly deleted someone's data because an invisible option in the UI was unknowingly toggled last week sounds bad.

> Maybe you mean that someone could inadvertently change the value when editing a cluster and making a mode selection? 

That is what I meant, if your cluster has been set (via YAML edit) to force=true, deleteLocalData=false, then as soon as you load this component the model is automatically changed to true/true.  Edit cluster -> save will overwrite their deleteLocalData=false with true.

And if the RKE template the cluster is based on says only one or the other is editable, then something weird happens (presumably they end up being out sync because your request to change a field that's not editable will be ignored on save.)

codyrancher:
@vincent99 I split the properties up instead of keeping the modes.

PR #4006, opened 2020-05-22
Removing the scope parameter from github oauth

codyrancher:
Types of changes
======

Linked Issues
======
rancher/rancher#27175


vincent99:
Looks fine, but needs the backend change in before merging or GitHub will be broken in the meantime

PR #4072, opened 2020-07-07
UI - Enabling desiredNodes field on the EKS driver cluster configuration screen

ribeiro-rodrigo:
This enabled change or the desired Nodes field in edit mode on the EKS cluster driver screen. Thus, it will be possible to change the number of nodes in a cluster through the Rancher graphical interface.

Types of changes

Linked Issues
https://github.com/rancher/rancher/issues/27847


deniseschannon:
As we are moving to the new EKS v2 support in Rancher v2.5, we encourage all users to upgrade and use the new provisioner which allows being able to change the number of nodes in EKS.

Thank you for your PR @ribeiro-rodrigo!

PR #4073, opened 2020-07-14
Handle mismatch between the region name and the region identifier for…

jlamillan:
Proposed changes
======

_Backport_ of [this PR](https://github.com/rancher/ui/pull/4069) to the 2.4 branch. (mismatch between the region name and the region identifier in OCI driver).

Types of changes
======

Linked Issues
======
[Here](https://github.com/rancher-plugins/rancher-machine-driver-oci/issues/2)
 

Further comments
======


jlamillan:
@westlywright

PR #4095, opened 2020-07-25
Add Azure German Public locations to the provider choices

adaliszk:
# Proposed changes
Adding German Public locations to the Azure choices

# Types of changes
New feature

# Linked Issues
https://github.com/rancher/rancher/issues/28082

# Comments
Check the infrastructure pages for double-checking the locations, marked with (Public) are in the `AzurePublicCloud` while the (Sovereign) is in the `AzureGermanCloud`.
- https://azure.microsoft.com/en-us/global-infrastructure/data-residency
- https://azure.microsoft.com/en-us/global-infrastructure/geographies

### DIFF
--- a/lib/shared/addon/utils/azure-choices.js
+++ b/lib/shared/addon/utils/azure-choices.js
@@ -41,6 +41,14 @@ export let regions = {
       'displayName': 'West Europe',
     },
     {
+      'name':        'germanynorth',
+      'displayName': 'Germany North',
+    },
+    {
+      'name':        'germanywescentral',
+      'displayName': 'Germany West Central',
+    },
+    {
       'name':        'japanwest',
       'displayName': 'Japan West',
     },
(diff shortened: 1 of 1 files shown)


lib/shared/addon/utils/azure-choices.js
```diff
@@ -40,6 +40,14 @@ export let regions = {
       'name':        'westeurope',
       'displayName': 'West Europe',
     },
+    {
+      'name':        'germanynorth',
+      'displayName': 'Germany North',
+    },
+    {
+      'name':        'germanywescentral',
```
turgutbaumann, line 48: Hi Ádám,
I'm not sure, but I guess you have a typo here, haven't you? I assume the region is name "germanywestcentral", so your commit is missing a "t".

Cheers
Turgut

adaliszk: Whops, nice catch!

PR #4099, opened 2020-07-28
Switching from min value to default value for upgradeStrategy timeout

codyrancher:
Proposed changes
======
Turns out that the min value that the backend accepts won't allow
upgrades to complete. This switches the value to the default value to
mitigate that issue.

Types of changes
======

Linked Issues
======
rancher/rancher#27333


codyrancher:
I updated the backend issue and mentioned the min value problem so they can fix that as well https://github.com/rancher/rancher/issues/28074#issuecomment-664706563.

PR #4133, opened 2020-08-14
Implementing max token ttl for api keys

codyrancher:
Proposed changes
======
- added 'auth-token-max-ttl-minutes' setting
- modified api-key modal to better handle the max-ttl when it is set

Types of changes
======

Linked Issues
======
rancherlabs/rancher-security#466

### auth-token-max-ttl-minutes > 0
- Max option
  - Show the largest unit that has a value > 1 otherwise default to minutes
- A custom number input with min and max values set
  - only show the units that can have a value >= 1
Screen Shot 2020-08-14 at 3 22 08 PM
Screen Shot 2020-08-14 at 3 46 38 PM

### auth-token-max-ttl-minutes <= 0
- Show what exists today
Screen Shot 2020-08-14 at 3 23 11 PM


codyrancher:
Waiting on approval https://github.com/rancherlabs/rancher-security/issues/466#issuecomment-674302869

PR #4156, opened 2020-09-02
Suppress exceptions when parsing JSON settings, default to raw value

nikkelma:
Proposed changes
======
In settings `danger-zone` component, suppress exceptions when parsing JSON settings, defaulting to raw value on any failure.

Types of changes
======
Bugfix

Linked Issues
======
https://github.com/rancher/rancher/issues/28504

Further comments
======
Resolves error missed by #4136, implements approach proposed in https://github.com/rancher/rancher/issues/28504#issuecomment-685967570.

### DIFF
--- a/lib/shared/addon/components/settings/danger-zone/component.js
+++ b/lib/shared/addon/components/settings/danger-zone/component.js
@@ -69,9 +69,17 @@ export default Component.extend({
       }
 
       if (get(details, 'kind') === 'json') {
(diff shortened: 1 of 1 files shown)


nikkelma:
Tested with a local dev server, confirmed to meet all known existing related errors. Will move this to a regular PR, if a different approach is wanted will happily refactor or let someone else take over.

PR #4169, opened 2020-09-14
color only theme changes

lvuch:
requested by darren


vincent99:
I don't think this is really what you're going for in light...

image

PR #4216, opened 2020-10-14
added 1.18 and 1.14 eks versions

paynejacob:
Added new eks version 1.18 and added back version 1.14 as it is not deprecated until december

https://github.com/rancher/rancher/issues/28317


paynejacob:
nvm this is for 2.4

paynejacob:
https://github.com/rancher/rancher/issues/29508

PR #4430, opened 2021-02-06
Changing how we reset values when switching credentials for vsphere node templates

codyrancher:
Proposed changes
======
Changing how we reset values when switching credentials for vsphere node templates

Types of changes
======

Linked Issues
======
rancher/rancher#30690


westlywright (approved):

westlywright (changes requested):
Pulled the PR down for review, I am seeing an error when I pull up the vSphere Node Template and I don't have any credentials. It is probably unrelated to your changes directly but can you take a look?

codyrancher:
Fixed two separate console errors that I saw.

1. vsphere credentials had a console error due to not passing in validation props to input-url so i added those props.
2. We were making requests for the vsphere options before getting datacenters so I prevent the requests from happening

westlywright (approved):

PR #4444, opened 2021-02-12
Provide more context around in-tree and out-of-tree cloud providers

codyrancher:
Proposed changes
======
Provide more context around in-tree and out-of-tree cloud providers

- Add helper text to help people find the out-of-tree providers for vsphere
- Label each in-tree provider with (In-tree) label
- Show a deprecated message for each in-tree option

Types of changes
======

Linked Issues
======
rancher/rancher#31162

image
image
image
image

### DIFF
--- a/lib/shared/addon/components/cru-cloud-provider/component.js
+++ b/lib/shared/addon/components/cru-cloud-provider/component.js
@@ -178,6 +178,20 @@ export default Component.extend({
     return out;
   }),
 
+  showVsphereHelperText: computed('selectedCloudProvider', 'driver', function() {
+    const driver = get(this, 'driver');
+    const selectedCloudProvider = get(this, 'selectedCloudProvider');
+
+    return (driver === 'custom' || driver === 'vmwarevsphere') &&
+           (selectedCloudProvider === 'external')
+  }),
+
+  isInTree: computed('selectedCloudProvider', function() {
+    const selectedCloudProvider = get(this, 'selectedCloudProvider');
+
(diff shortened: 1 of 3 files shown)


westlywright (approved):

sadiapoddar:
Verified on 2.5-head, 03b8153
Test1 
1. Selected custom (in-tree) provider and also noticed the in-tree deprecation warning and message. 
2. provided the cloud provider config in cluster YMAL and
3. Created the cluster as in-tree

Test2
1. Selected "external" for creating an out-of-tree and noticed a warning which is not relevant. Instead we can say the requirement as "After creating an external cluster, install cpi and csi driver to make out-of-tree  ...... " 
I am not sure about the exact wording but I listed the workflow below, we need to communicate that with the users.

Screen Shot 2021-02-18 at 6 14 49 PM

2. The links for CPI and CSI driver are misleading during creating a cluster, these driver needs to be installed once the external cluster is created and ready. Also, there is some dependency between these drivers. CPI needs to installed first, after installation is complete, we need to check if all the nodes are marked with the vSphere provider information before we can start installing CSI driver. 

**The right workflow for the external (out-of-tree) are as follows**

1. create a cluster by selecting external, 
Screen Shot 2021-02-18 at 6 06 07 PM

2. Once the cluster is up, install CPI driver 
3. Chek all the nodes are marked with the vSphere provider information
```
kubectl describe nodes | grep "ProviderID"
```
4. Then install CSI driver


Reopening this bug, we need to remove the CPI and CSI driver link from the cluster creation page and change the warning message.

PR #4467, opened 2021-02-23
Adding the agent env vars to rke drivers.

codyrancher:
Proposed changes
======
Adding the agent env vars to rke drivers.

Types of changes
======

Linked Issues
======
rancher/rancher#31023

image
image

### DIFF
--- a/app/authenticated/cluster/edit/route.js
+++ b/app/authenticated/cluster/edit/route.js
@@ -9,6 +9,7 @@ import { scheduleOnce } from '@ember/runloop';
 export default Route.extend({
   access:                 service(),
   globalStore:            service(),
+  clusterStore:           service(),
   releaseVersions:        service(),
   clusterTemplateService: service('clusterTemplates'),
   roleTemplateService:    service('roleTemplate'),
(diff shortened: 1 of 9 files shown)


codyrancher:
@westlywright just a note about rke templates. I disabled this for rke templates because the templates really don't seem to have great support for arrays/objects. I attempted to kludge it in but ended up with a worse mess. If it's something we want to be able to template I think we should put it into a separate issue as it adds a lot of scope.

westlywright (changes requested):
Didn't find anything specific in the code but when running it, when I edit a fresh linode cluster and select the secret type I get an undefined error.

westlywright:
@codyrancher looks like a minor bug with the `Secret` type. Also I think that not doing the templates is okay for now. Can you comment on the ticket so it is clear to whomever may wonder?

codyrancher:
Disabling secrets and configmaps if they aren't available. This also simplified the interface of FormAgentEnvVar by getting rid of the showSecrets/ConfigMap props.

image

westlywright (approved):

PR #4481, opened 2021-03-01
Switching over to having the envVars just be keyValues

codyrancher:
Proposed changes
======
After some discussions we decided to simplify things since all that's actually needed are keyValues.

Types of changes
======

Linked Issues
======
rancher/rancher#31545

image


codyrancher:
Just as a note. I left the order as Advanced then K3s options because the change involved would required pushing this into driver-import as well as leaving it in cru-cluster which I think is risky at this stage. I'll update the ticket to inform QA.

westlywright (approved):

PR #4544, opened 2021-04-15
Updating the principal logicalType to account for okta

codyrancher:
Proposed changes
======
Updating the principal logicalType to account for okta

This will allow the group principals to be displays on the project members section.

Types of changes
======

Linked Issues
======
rancher/rancher#30401

image

### DIFF
--- a/app/models/principal.js
+++ b/app/models/principal.js
@@ -52,6 +52,7 @@ var Principal = Resource.extend({
     case C.PROJECT.TYPE_PING_USER:
     case C.PROJECT.TYPE_RANCHER:
     case C.PROJECT.TYPE_SHIBBOLETH_USER:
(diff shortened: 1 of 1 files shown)


codyrancher:
@westlywright out of curiosity, do  you know why we limit what we show for the project members to groups? I thought that was odd.

vincent99:
The list in the dropdown is "groups that I am a member of", which is what `GET /v3/principals` with no search query returns (plus the user principal for yourself).

In many providers you couldn't enumerate all the possible users even if you wanted to (e.g. GitHub has millions) or there's no way to search/list them at all (SAML), so this is pretty much the only useful option we have to offer.

westlywright (approved):

PR #4629, opened 2021-05-20
Fixing how the frontend handles the bracket annotation when building cluster answers

codyrancher:
Proposed changes
======
The front end wasn't sending answers to the backend when a user set the the
privateRegistries field when using an RKE template which allowed the field to
be overriden. This was happening because we weren't handling the bracket
notation properly.

Types of changes
======

Linked Issues
======
Front End issue: https://github.com/rancher/dashboard/issues/3940
Back End Issue: https://github.com/rancher/rancher/issues/27166

Before the fix
======
private-registry-broken

After the fix
======
private-registry-after

### DIFF
--- a/lib/shared/addon/components/cluster-driver/driver-rke/component.js
+++ b/lib/shared/addon/components/cluster-driver/driver-rke/component.js
@@ -1014,7 +1014,11 @@ export default InputTextFile.extend(ManageLabels, ClusterDriver, {
 
     if (questions && questions.length) {
       questions.forEach((quest) => {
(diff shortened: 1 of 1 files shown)


jordojordo (approved):

codyrancher:
@jordojordo if you want to test this you'll need to make sure you have the latest version of rancher to get the backend changes.

jordojordo:
@codyrancher I tested it with `v2.6-1b28f3f6579e20eca8aac993a1b229f9e6d0a8cd-head` and saw the updated values saved correctly.

PR #4671, opened 2021-06-16
Avoid unintentional version upgrade (fix #33217)

barla95:
Proposed changes
======

When going to upgrade an app the template version has the last version preselected. If someone is not carefull enough and doesn't check the version each time could upgrade the app version unintentionally. This is a problem moslty when you want to change only the answer 

Types of changes
======

Bugfix

Linked Issues
======
 
https://github.com/rancher/rancher/issues/33217

Further comments
======


westlywright:
@codyrancher can you review this with QA? I seem to remember this being an intentional choice but the feature was written so long ago I can't find the original requirements.

nwmac:
Going to close out, as as-designed - please re-open if this is still something you feel is important.

PR #4692, opened 2021-07-01
[2.6.x] GKE V2 Node Tags

westlywright:
Proposed changes
======
Exposes the new tags param on the gke node pool config.

Types of changes
======
New Feature

Linked Issues
======
rancher/rancher#33416
 

Further comments
======
Should be held until the back-end is in


snasovich:
Duplicating comment from the issue for visibility here:
Please note that Tags can only be set on create, so they should be read-only on edit forms.
https://github.com/rancher/gke-operator/pull/60#issuecomment-872665710

westlywright:
@snasovich this should already be the case in the UI, thanks for adding that comment.

codyrancher (approved):

PR #4697, opened 2021-07-06
Adding K8S 1.20 as an available version for EKS

codyrancher:


codyrancher:
https://github.com/rancher/rancher/issues/33475

PR #4757, opened 2021-09-16
Update the name for the Packet node driver to Equinix Metal

rak-phillip:
Proposed changes
======
This updates labels for the Packet node driver to Equinix Metal. I understand that this change only requires updates to the labels and that the actual name of the driver will stay the same. 

Types of changes
======
Updates to English translations.

Linked Issues
======
rancher/dashboard#4161
rancher/rancher#33758

Further comments
======


codyrancher (approved):

vincent99:
There is a newer driver which has been renamed and we should probably switch to on the backend, but this is fine for now until they add it.

PR #4765, opened 2021-09-23
[master] Update initial kubernetes version for AKS

rak-phillip:
Proposed changes
======
This updates the initial kubernetes version from the string '1.19.9' to an actual value in the list of versions.

Types of changes
======

Bugfix for rancher/dashboard#4262 - Kubernetes version defaults to 1.21.2 in Cluster Options and 19.9.9 in Node Pools

Linked Issues
======
 
rancher/dashboard#4262

Further comments
======

### DIFF
--- a/lib/shared/addon/components/cluster-driver/driver-aks/component.js
+++ b/lib/shared/addon/components/cluster-driver/driver-aks/component.js
@@ -270,8 +270,9 @@ export default Component.extend(ClusterDriver, {
 
         const isEdit                        = mode === 'edit';
         const versionz                      = (get(versions, 'body') || []);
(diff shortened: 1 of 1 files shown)


rak-phillip:
I guess I should mention that I was looking at making use of the original solution that was commented out, `Semver.maxSatisfying(versionz, this.defaultK8sVersionRange)`, but this is throwing an error ` TypeError: xhr.body is undefined`

I'm guessing that's why the original solution was commented out in the first place and probably still needs fixing. I'm still taking a look to see if I can get that to work, but I thought I'd provide a little more background on why I chose the solution that I did

PR #4775, opened 2021-10-10
Add GCP asia-southeast2 (Jakarta) region

rawis:
Please Add GCP asia-southeast2 (Jakarta) region to register / import GKE Cluster


nwmac (approved):
LGTM

PR #4783, opened 2021-11-25
Use default catalog template versions

catherineluse:
This PR addresses both of these related issues: https://github.com/rancher/dashboard/issues/4627 and https://github.com/rancher/dashboard/issues/4628

Another related issue is https://github.com/rancher/dashboard/issues/4645

This PR makes it so that if you have a legacy multicluster app where the max Rancher version is v2.6.2, then you upgrade Rancher to v2.6.3, then try to upgrade the app, you can at least see the upgrade form now. Previously it was just erroring out.

But now if you make a change to the app template and click upgrade, you get an error from the back end saying the version is incompatible. I think this means the back end needs another change to allow legacy apps to be updated even if their chart version is incompatible with Rancher.

This screenshot shows an app installed in v2.6.2, with the Rancher max version set to v2.6.2, which now still has an upgrade form in v2.6.3rc1:
Screen Shot 2021-11-25 at 8 38 37 PM

But there is some additional back end work to be done because when you try to upgrade it, the back end doesn't allow it:
Screen Shot 2021-11-25 at 8 38 22 PM

The PR also makes it so that when you go to find the installed multi-cluster app in Apps & Marketplace, if you go to upgrade it, instead of showing you a broken form, it gives you a link back to the multi-cluster app page with the form that works:
Screen Shot 2021-11-26 at 3 41 16 PM
The result is:
Screen Shot 2021-11-26 at 3 41 28 PM

### DIFF
--- a/app/models/multiclusterapp.js
+++ b/app/models/multiclusterapp.js
@@ -76,7 +76,15 @@ const MultiClusterApp = Resource.extend({
     upgrade() {
       const templateId    = get(this, 'externalIdInfo.templateId');
       const catalogId     = get(this, 'externalIdInfo.catalog');
-      const vKeys         = Object.keys(get(this, 'catalogTemplate.versionLinks'));
+      let vKeys         = Object.keys(get(this, 'catalogTemplate.versionLinks'));
+
+      if ( !vKeys || !Object.keys(vKeys).length ) {
+        // If you install a legacy app and then upgrade Rancher,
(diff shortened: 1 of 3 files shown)


codyrancher (approved):

catherineluse:
This PR would not be needed when this issue is done https://github.com/rancher/rancher/issues/35680.

PR #4791, opened 2021-12-16
Use currentVersion parameter to upgrade legacy apps

catherineluse:
This PR partly addresses https://github.com/rancher/dashboard/issues/4774 by using the `currentVersion` parameter when requesting template details in the form to upgrade a legacy app.

The currentVersion parameter takes advantage of this fix https://github.com/rancher/rancher/issues/35680 which made it so that the available `versionLinks` are not empty when you try to upgrade a legacy app after upgrading Rancher to a version that is no longer supported by the legacy app.

To test this PR,

1. In Rancher v2.6.2, I enabled the legacy feature flag and installed legacy apps in the local `system` project.
2. I upgraded Rancher to v2.6.3-rc5 (because v2.6-head is considered prior to v2.6.0 for version compatibility)
3. I went to the legacy apps in the system project and tried to upgrade them
4. I confirmed that the form rendered successfully.
Screen Shot 2021-12-16 at 12 23 38 AM

Screen Shot 2021-12-16 at 12 48 55 AM

Caveats

Previously, when trying to solve the problem of the form not rendering when `versionLinks` was empty, I created a PR https://github.com/rancher/ui/pull/4783 that replaced `versionLinks` with an object containing the current app version, assuming the current version was valid. This was not an ideal solution because the current app version on the front end was not guaranteed to be valid on the back end. It was also not a complete solution because even though adding the current version to `versionLinks` was enough to make the form render, the result of actually submitting the form was a back end validation error because the back end indicated the version was incompatible, even when the upgrade contained minor changes that did not involve changing the app version.

This PR is also incomplete for the same reason, in that the back end still blocks form submission:

```
Validation failed in API: incompatible kubernetes version [1.21.7+k3s1] for template [system-library-rancher-logging-0.3.0]
```
Screen Shot 2021-12-16 at 12 23 22 AM

There still needs to be work done for the back end to allow the change. The failing upgrade request is to:

```
<server URL>/v3/project/local:p-gw9l2/apps/p-gw9l2:rancher-logging?action=upgrade
```

and the request payload is:

```
{"externalId":"catalog://?catalog=system-library&template=rancher-logging&version=0.3.0",
"answers":{"fluentd.cluster.dockerRoot":"/var/lib/docker","fluentd.enabled":"true","fluentd.fluentd-linux.cluster.dockerRoot":"/var/lib/docker","fluentd.fluentd-linux.enabled":"true","fluentd.fluentd-windows.enabled":"false","log-aggregator.enabled":"true","log-aggregator.flexVolumeDir":"/usr/libexec/kubernetes/kubelet-plugins/volume/exec","log-aggregator.log-aggregator-linux.enabled":"true","log-aggregator.log-aggregator-linux.flexVolumeDir":"/usr/libexec/kubernetes/kubelet-plugins/volume/exec"},
"valuesYaml":"",
"forceUpgrade":false}
```

Technically the back end already has the current app version because it is in the `externalId` in the request payload. I believe will still need to change to allow that request to go through.

### DIFF
--- a/app/catalog-tab/launch/route.js
+++ b/app/catalog-tab/launch/route.js
@@ -21,18 +21,29 @@ export default Route.extend({
   model(params, transition) {
     const { store, clusterStore } = this;
 
-    const dependencies = {
-      tpl:        get(this, 'catalog').fetchTemplate(params.template),
-      namespaces: clusterStore.findAll('namespace')
-    };
-
-    if ( params.upgrade ) {
-      dependencies.upgrade = get(this, 'catalog').fetchTemplate(`${ params.template }-${ params.upgrade }`, true);
-    }
+    let dependencies = { namespaces: clusterStore.findAll('namespace') };
 
     if (params.appId) {
       dependencies.app = store.find('app', params.appId);
+      dependencies.app.then((appData) => {
+        const getCurrentVersion = (app) => {
+          const externalId = app.externalId;
+          const splitId = externalId.split('version=')
+          const currentVersion = splitId[1];
+
+          return currentVersion;
+        }
+        const currentVersion = getCurrentVersion(appData);
+
+        dependencies.upgrade = get(this, 'catalog').fetchTemplate(`${ params.template }-${ params.upgrade }`, true, currentVersion);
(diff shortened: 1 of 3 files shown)


nwmac (approved):
LGTM

Validated locally with upgrade

PR #4797, opened 2022-01-05
Only allow PNI to be selected if a network policy is selected

codyrancher:
Proposed changes
======
Only allow PNI to be selected if a network policy is selected

Types of changes
======

Linked Issues
======
rancher/dashboard#4452
pni


jordojordo (approved):
lgtm

PR #4804, opened 2022-01-13
Fix issue with constantly requested catalog image

nwmac:
Addresses https://github.com/rancher/dashboard/issues/4803

Only an issue in production.

You can run a production build locally with:

`./node_modules/.bin/ember serve --environment production`

This PR uses the LazyIcon mixin that is used elsewhere.


nwmac:
@richard-cox Can you remind me we need to review this for 2.6.4.

richard-cox (approved):

PR #4815, opened 2022-02-24
Can enable project network isolation for imported clusters

catherineluse:
This PR addresses https://github.com/rancher/dashboard/issues/4574 by adding radio buttons to allow enabling project network isolation for imported clusters.

Screen Shot 2022-02-24 at 3 04 05 PM

Currently this input is going to always be displayed because for imported clusters, we don't know if the cluster has a CNI that supports project network isolation. (Ryan Sanna to confirm.)

When PNI is enabled, we show the same warning that is used when the option is enabled for RKE1 clusters:
Screen Shot 2022-02-24 at 2 44 42 PM

## Testing

To test this PR,

1. I imported a K3s cluster (I tested the same steps on both K3s and RKE2 clusters)
2. In Cluster Management, I went to the imported cluster and clicked Edit Config
3. Went to Project Network Isolation and clicked Enabled
4. Clicked Save

Verified that for both K3s and RKE2, `enabledNetworkPolicy` was set to true in the cluster data in the network request
Screen Shot 2022-02-24 at 5 20 47 PM

## Outstanding Questions

Which Kubernetes distro(s) do we want to support? On both K3s and RKE2 clusters, when you save the changes, you get the API error that says it is not a valid option to set `enableNetworkPolicy` to true:
Screen Shot 2022-02-24 at 5 20 09 PM

Also, Cody noticed that the form for K3s clusters already exposes an option for enabling PNI under advanced options, even though the API throws the above error if you enable it. Should this option be exposed under Advanced Options for both K3s and RKE2? Currently the form for RKE2 doesn't have an advanced options section.

Here are the existing advanced options for imported K3s clusters:
Screen Shot 2022-02-24 at 4 36 31 PM

### DIFF
--- a/app/styles/components/_accordion.scss
+++ b/app/styles/components/_accordion.scss
@@ -98,3 +98,7 @@ $acc-label              : $text-muted !default;
     display: inline-block;
   }
 }
+
+.reservedSpaceForBanner {
+  height: 80px;
+}
--- a/lib/shared/addon/components/managed-import-cluster-info/component.js
+++ b/lib/shared/addon/components/managed-import-cluster-info/component.js
@@ -43,6 +43,17 @@ export default Component.extend({
     return get(this.config, upgradeStrategyPath);
   }),
 
+  enableNetworkPolicy: computed('config', 'configField', {
+    get() {
+      return !!get(this.config, 'enableNetworkPolicy');
+    },
+    set(key, value) {
+      set(this, 'cluster.enableNetworkPolicy', value);
+
+      return value;
+    }
+  }),
+
   serverConcurrency: computed('upgradeStrategy.serverConcurrency', {
     get() {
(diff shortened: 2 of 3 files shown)


nwmac:
@catherineluse Is this still draft and needed for 2.6.4?

catherineluse:
Yes, it's still in draft because this PR adds it to RKE2 clusters. I assumed this was one of the RKE2 cluster provisioning features but it's not. Ryan clarified that this change affects RKE1, AKS and GKE clusters only, so I'll need to change it to affect those clusters.

Also the K3s has a PNI option but that should actually be removed because the feature is not supported for K3s. I was going to take care of that in the same PR as well.

As to whether the feature is needed for v2.6.4, I would say yes because the backend part is still in the v2.6.4 milestone.

catherineluse:
Closing as stale

PR #4821, opened 2022-03-16
Set max-width to the form to avoid overflow of their own components

cnotv:
#5414

Prevented `form` content to overflow in some cases by setting max width, e.g. RKE1 Node Template cloud-config field for Sphere.

Screenshot from 2022-03-16 19-10-41


rak-phillip (approved):
LGTM, thanks for scooping this up 😁

PR #4822, opened 2022-03-17
Use existing cluster config as defaults for RKE template clusters

catherineluse:
This PR addresses https://github.com/rancher/dashboard/issues/5032 by making it so that if you create a cluster from an RKE template and override some values, the next time you edit the form, your own values are used as the defaults instead of the defaults for provisioning a new cluster.

To test this PR:

1. I created an RKE template and set the 'Node Port Range' to `30000-32767`. Set 'Allow user override' for the Node Port Range.
2. Created an RKE1 cluster from the RKE template. Overrode the value for node port range and set it to `30000-31000`.
3. In cluster manager, went to the new cluster and clicked **Edit Config.**
4. Confirmed that my override value - `30000-31000` - was shown as the default instead of the default value from the RKE template.
Screen Shot 2022-03-16 at 5 07 34 PM

Related issue: https://github.com/rancher/rancher/issues/21758

### DIFF
--- a/lib/shared/addon/components/cluster-driver/driver-rke/component.js
+++ b/lib/shared/addon/components/cluster-driver/driver-rke/component.js
@@ -1654,12 +1654,13 @@ export default InputTextFile.extend(ManageLabels, ClusterDriver, {
   },
 
   initClusterTemplateQuestions() {
-    let {
-      clusterTemplateQuestions,
-      primaryResource,
-    } = this;
+    let { clusterTemplateQuestions } = this;
 
-    if (clusterTemplateQuestions && clusterTemplateQuestions.length > 0) {
+    // Use the existing cluster config values as defaults for the questions
+    // because one or more values could already be overridden.
+    let primaryResource = get(this, 'config');
+
+    if (clusterTemplateQuestions && clusterTemplateQuestions.length > 0 && primaryResource) {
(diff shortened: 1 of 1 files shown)


codyrancher:
@catherineluse did you verify that this works for other fields and for editing clusters that do and don't have questions? I'm a little reluctant to have this one merged since it's changing the generic implementation of questions for what appears to be a problem for a single field.

catherineluse:
@codyrancher I tried it with other fields and it seemed to work, but I'm closing this PR because I think it's the wrong approach. I think I was wrong about the root cause - it's not ember that had a problem, but the dashboard. I realized when working on https://github.com/rancher/dashboard/issues/5395 that when you edit a cluster that was created from an RKE template in the Ember UI, the URL has the RKE template version in it, but the template version wasn't being passed in by the dashboard to the iFramed ember page. I fixed that issue by passing in the RKE template version as a param to the ember page and it looks like that change solves this issue as well.

catherineluse:
Reopening because Nancy proved the problem was with Ember, not the dashboard

mantis-toboggan-md:
LGTM - tested overriding template values and editing a non-template cluster in Ember and Ember-in-Vue

mantis-toboggan-md (approved):

PR #4824, opened 2022-04-01
reinitialize node pools and members when cluster changes

mantis-toboggan-md:
Proposed changes
======
This PR updates the cluster edit/create node pool and member components to re-initialize pools and members respectively when the cluster changes. This bug appears in the iFramed UI running in the dashboard, but not in the Ember UI running in standalone. It seems that when using the dashboard to navigate between editing RKE1 clusters, the embedded Ember UI navigates directly between the two clusters' edit views, so any relevant components' `init` function isn't re-run, and consequently the node pools and cluster members displayed aren't updated to reflect the new cluster.

Linked Issues
======

https://github.com/rancher/dashboard/issues/5343 - in addition to this problem with cluster members, you can see that the node pools section also shows the previous RKE1 cluster's config.

### DIFF
--- a/lib/shared/addon/components/cru-node-pools/component.js
+++ b/lib/shared/addon/components/cru-node-pools/component.js
@@ -97,24 +97,7 @@ export default Component.extend({
 
   init() {
     this._super(...arguments);
-    const originalPools = (get(this, 'cluster.nodePools') || []).slice();
-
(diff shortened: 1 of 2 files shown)


catherineluse (approved):
LGTM

PR #4840, opened 2022-05-05
Add dispatch step to update tag in rancher/rancher

superseb:
Follow-up to https://github.com/rancher/ui/pull/4835 which was reverted because it broke the build. I think because this step introduces `depends_on`, it causes the other steps to be executed in parallel causing ordering problems. This PR has the same changes without `depends_on`.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -113,6 +113,24 @@ steps:
       include:
       - "refs/tags/v2.6.*"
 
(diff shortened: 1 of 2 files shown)


superseb:
@richard-cox can we get this merged at your earliest convenience to see if it not messes up the builds and we are ready for when we are going to create tags for 2.6.6

PR #4854, opened 2022-06-16
fix: Only display valid Linode types for LKE node pools

LBGarber:
This pull request alters the Linode LKE node pool type filtering logic to only include valid instance types. Previously the filter would allow certain invalid types and exclude valid types (`highmem`).

See: https://www.linode.com/docs/api/linode-types/#type-view__responses

Related issue: https://github.com/rancher/dashboard/issues/6424


LBGarber:
https://github.com/rancher/rancher/issues/38119

PR #4893, opened 2022-09-16
Fix select option disabled color

nwmac:
On Linux, disabled options in the default OS select drop down can not be distinguished from normal ones.

This PR fixes this.

See: https://github.com/rancher/dashboard/issues/6774 for an example


nwmac:
Closed in favour of #4894

PR #4905, opened 2022-10-21
Add StandardSSD as a storage option for Azure cluster provisioning

catherineluse:
This PR partly addresses https://github.com/rancher/dashboard/issues/7177 by adding StandardSSD to Azure storage options.

To test this PR,

1. Went to Cluster Management > RKE1 Configuration > Node Templates
2. Clicked Add Template
3. Clicked Azure
4. Went to the Instance section
5. Went to the Storage Type dropdown menu
6. Selected Standard SSD LRS, the new storage option

Screen Shot 2022-10-27 at 6 09 20 PM

8. Added a name for the node template
9. Clicked Create
10. Created an RKE1 Kubernetes cluster in Azure using the new node template
11. Confirm that `StandardSSD_LRS` is sent in the network request

### DIFF
--- a/lib/shared/addon/utils/azure-choices.js
+++ b/lib/shared/addon/utils/azure-choices.js
@@ -454,6 +454,10 @@ export let storageTypes = [
   {
     name:  'Premium LRS',
     value: 'Premium_LRS',
+  },
+  {
+    name:  'Standard SSD LRS',
+    value: 'StandardSSD_LRS'
   }
 ];
 


HarrisonWAffel:
@catherineluse I tested this using the `rancher-machine` binary, and it looks like the `StandardSSD_LRS` option is only supported if you enable managed disks in the node template. I found this documentation https://learn.microsoft.com/en-us/azure/virtual-machines/faq-for-disks#can-i-use-standard-ssds-as-unmanaged-disks- which confirms this.

catherineluse:
I can add this restriction to the UI:

- when StandardSSD_LRS is selected, the managed disk dropdown becomes disabled and its value becomes enabled
- managed disk dropdown has tooltip that says StandardSSD_LRS requires managed disk to be enabled

mantis-toboggan-md (approved):

PR #4909, opened 2022-10-26
Current GKE/AKS version is shown even if below min version

catherineluse:
This PR fixes https://github.com/rancher/dashboard/issues/7276 by changing the logic for hiding unsupported GKE and AKS versions.

Previously, K8s versions 1.22 and below were always hidden. This change creates an exception for the edit form if the AKS or GKE cluster is already 1.22, in which case it shows the current version as the default.


mantis-toboggan-md (approved):
LGTM

PR #4911, opened 2022-10-28
removed calls to logging apis while workload creation

vardhaman22:
Proposed changes
======

fix for workload creation after v1 Logging removal from the server.

Types of changes
======

Bugfix

Linked Issues
======
 
https://github.com/rancher/rancher/issues/39350

Further comments
======


nwmac:
Closing in favour of #4912 which also removes logging from nav

PR #4914, opened 2022-11-07
Fix image option of Harvester Node Template

masap:
Proposed changes
======
I tried to create Node Template of a imported Harvester via Rancher UI. I could see Namespace options and Network Name options. But I could not see any Image option even when I have uploaded iso image on Harvester. This PR fixes image option of Harvester Node Template.

Types of changes
======

Linked Issues
======
SURE-5517

Further comments
======

Signed-off-by: Masashi Honma <masashi.honma@gmail.com>


masap:
Screen shot of this issue.

image-not-found

masap:
I recognized iso image should not be used.

PR #4918, opened 2022-11-10
Adding us-west4 GCP region

pramodsetlur:
Proposed changes
======
This PR is adding the us-west4 GCP region into the UI’s config.

Types of changes
======
This technically seems like a new non-breaking feature request.

Linked Issues
======
https://github.com/rancher/rancher/issues/39571

Further comments
======


nwmac:
GKE support was migrated to Vue, so this is no longer needed.

PR #4910, opened 2022-10-27
Can add Azure tags to RKE1 node template

catherineluse:
https://github.com/rancher/dashboard/issues/7164

This PR adds the "Azure Tags" button for RKE1 Azure node templates:
Screen Shot 2022-10-26 at 9 12 39 PM

Screen Shot 2022-10-26 at 9 12 49 PM

PR #4924, opened 2022-12-13
Revert "Can add Azure tags to RKE1 node template"

catherineluse:
Reverts rancher/ui#4910


catherineluse:
I am closing the PR to revert the feature because it is being kept in the release after all

PR #4925, opened 2022-12-14
RKE1 cluster create/edit forms no longer throw 404s

catherineluse:
This PR fixes three bugs:

## Creating a custom rke1 cluster switches UI to cluster list view instead of create/edit form
https://github.com/rancher/dashboard/issues/7699

To test, I went to create an RKE1 custom cluster and confirmed that the RKE1 custom cluster creation form appears instead of the old Ember cluster list view.

## Editing an active rancher provisioned rke1 cluster navigates to ember UI page
https://github.com/rancher/dashboard/issues/7688
To test, I created an RKE1 cluster, went to Edit Config, and confirmed that the cluster edit form appears instead of the old Ember cluster list view.

## vSphere Node Driver cluster goes back to old UI
https://github.com/rancher/dashboard/issues/7636

To test, I went to create all types of RKE1 node driver clusters, including vSphere, and confirmed that the correct form appears instead of the old Ember cluster list view.

### DIFF
--- a/app/authenticated/cluster/route.js
+++ b/app/authenticated/cluster/route.js
@@ -30,7 +30,6 @@ export default Route.extend(Preload, {
         return get(this, 'scope').startSwitchToCluster(cluster).then(() => {
           if ( get(cluster, 'isReady') ) {
             const preloads = [
-              this.preload('clusterScan', 'globalStore'),
               this.preload('namespace', 'clusterStore'),
               this.preload('persistentVolume', 'clusterStore'),
               this.preload('storageClass', 'clusterStore'),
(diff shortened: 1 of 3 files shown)


richard-cox (approved):
I found another reference to `clusterScan` (`app/authenticated/cluster/route.js` `this.preload('clusterScan', 'globalStore'`) that throws a 404, though as far as i can tell that's only used in unsupported cluster detail pages (and i didn't see any loss of functionality there).

This also needs backporting to 2.6, though no branch for that exists in `rancher/ui`. I'll create one today

catherineluse:
I've removed the second preload for clusterscan

codyrancher (approved):

PR #4932, opened 2022-12-19
RKE1 Azure node templates have accelerated networking and availability zones

catherineluse:
Addresses https://github.com/rancher/dashboard/issues/7753 and https://github.com/rancher/dashboard/issues/7752

This PR is functionally the same as https://github.com/rancher/dashboard/pull/7687, but for RKE1 node templates.

### New options
- Can select an availability zone.
- Can enable accelerated networking.
- The VM size dropdown is grouped so that you can see which VMs support accelerated networking and which do not.

### Refetching
- VMs are refetched when the region changes because different sizes are available per region.

### Error messages
- Warn the user when the selected VM size is not in the selected region.
- Warn the user if they have selected availability zones, but that is not supported for the region or VM size.
	- For example, South Africa North has availability zones for some sizes but not others. So `Standard_A2_v2` in that region has three AZs and that works, but if you change the size to `Standard_A2_v2` you see an error.
- Warn the user if they wanted availability zones, but availability zones are not supported in the selected region.
- Warn the user if accelerated networking is enabled but the selected VM size does not support accelerated networking. Show the warning by the vm sizes and by the AN checkbox.

### Notes
- Replaced the hardcoded VM size list with V2 API call to get VM sizes.
- Replaced the hardcoded region list with the API call to get regions, to show the full list and be consistent with AKS and RKE2/K3s provisioning.
- I deleted the "environment" choice dropdown because it was last updated five years ago, and while I assume it worked at the time, it doesn't work like that anymore. Now the environment, such as Azure public cloud or the government cloud, is configured in the cloud credentials, and the credentials are used to fetch the regions, so the environment should not be able to be configured in this form. If you want to create nodes in a different environment, you should use a different cloud credential. We may want to call that out in the release notes.
- Because the form now needs to call the Azure API with the user's credentials, I have changed the form so that you can't see it until valid credentials are selected.
- I think there is a backend bug where if you first create a node template with an availability zone, then change it to an availability set, the change does not get persisted. Edit: Fixed by clearing out values using empty string instead of null.

### DIFF
--- a/lib/nodes/addon/components/driver-azure/component.js
+++ b/lib/nodes/addon/components/driver-azure/component.js
@@ -175,9 +175,9 @@ export default Component.extend(NodeDriver, {
     // If the user switches between availability sets and availability zones,
     // the other value should be cleared out.
     if (get(this, 'useAvailabilitySet')) {
-      set(this, 'config.availabilityZone', null);
+      set(this, 'config.availabilityZone', '');
     } else {
-      set(this, 'config.availabilitySet', null);
+      set(this, 'config.availabilitySet', '');
     }
   }),
 
--- a/lib/nodes/addon/components/driver-azure/component.js
+++ b/lib/nodes/addon/components/driver-azure/component.js
@@ -246,7 +246,7 @@ export default Component.extend(NodeDriver, {
     const withAnLabel = intl.t('nodeDriver.azure.size.supportsAcceleratedNetworking');
     const vmsWithAn = get(this, 'vmsWithAcceleratedNetworking');
     const vmsWithoutAn = get(this, 'vmsWithoutAcceleratedNetworking');
-    const out = [{
+    let out = [{
       kind:     'group',
       name:     noAnLabel,
       value:    noAnLabel,
@@ -261,15 +261,7 @@ export default Component.extend(NodeDriver, {
       })
       .concat(vmsWithAn)
 
-    if (!get(this, 'selectedVmSizeExistsInSelectedRegion')) {
-      out.push({
-        label:     get(this, 'config.size'),
-        value:    get(this, 'config.size'),
-        disabled: true
-      });
-    }
-
-    return out.map((vmData) => {
+    out = out.map((vmData) => {
       const { Name } = vmData;
 
       if (vmData.kind === 'group') {
@@ -282,6 +274,16 @@ export default Component.extend(NodeDriver, {
         disabled: vmData.disabled || false,
       };
     });
+
+    if (!get(this, 'selectedVmSizeExistsInSelectedRegion')) {
+      return out.concat({
+        label:    get(this, 'config.size'),
+        value:    get(this, 'config.size'),
+        disabled: true
+      });
+    }
+
+    return out
   }),
   privateSet: computed('publicIpChoice', 'publicIpChoices', function() {
     let publicIpChoice = get(this, 'publicIpChoice');
@@ -349,12 +351,9 @@ export default Component.extend(NodeDriver, {
       // Don't show the warning until the VM sizes are loaded
       return '';
     }
-    const selectedVmIsAvailableInSelectedRegion = this.vmSizes.filter((vmData) => {
-      return vmData.Name === get(this, 'config.size');
-    }).length > 0;
     const intl = window.l('service:intl')
 
-    if (!selectedVmIsAvailableInSelectedRegion) {
+    if (!get(this, 'selectedVmSizeExistsInSelectedRegion')) {
       return intl.t('nodeDriver.azure.size.availabilityWarning');
     }
 
--- a/lib/nodes/addon/components/driver-azure/component.js
+++ b/lib/nodes/addon/components/driver-azure/component.js
@@ -6,9 +6,11 @@ import { scheduleOnce } from '@ember/runloop';
 import Component from '@ember/component';
 import NodeDriver from 'shared/mixins/node-driver';
 import layout from './template';
-import { regions, sizes, storageTypes, environments } from 'ui/utils/azure-choices';
+import { storageTypes, environments } from 'ui/utils/azure-choices';
 import { inject as service } from '@ember/service';
 import { randomStr } from 'shared/utils/util';
+import { addQueryParams } from 'shared/utils/util';
+import { hash } from 'rsvp';
 
 const DRIVER = 'azure';
 const CONFIG = 'azureConfig';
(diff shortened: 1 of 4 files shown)


catherineluse:
I fixed the issue where switching from an AZ to an availability set wasn't being persisted - I don't intend to make any more changes to the PR before it's reviewed

lib/nodes/addon/components/driver-azure/template.hbs
```diff
@@ -264,13 +376,35 @@
...
+            optionLabelPath="value"
+            optionGroupPath="group"
+            value=config.size
+          }}
+          {{#if showVmSizeAvailabilityWarning}}
+            {{#banner-message color="bg-error"}}
+              <p>{{vmSizeAvailabilityWarning}}</p>
+            {{/banner-message}}
+          {{/if}}
+          {{#if showVmAvailabilityZoneWarning}}
```
mantis-toboggan-md, line 394: I'm not sure it always makes sense to show `vmSizeAvailabilityWarning` here, since one of the two potential error messages is asking the user to re-configure one of two options in a different section. `regionSupportsAzsButNotThisSize` certainly makes sense here, but probably not `regionDoesNotSupportAzs` :

Screen Shot 2022-12-19 at 2 06 00 PM

lib/nodes/addon/components/driver-azure/component.js
```diff
@@ -165,6 +303,151 @@ export default Component.extend(NodeDriver, {
...
+    if (get(this, 'regions.length') === 0) {
+      // Don't show the warning until the regions are loaded
+      return '';
+    }
+    if (get(this, 'vmSizes.length') === 0) {
+      // Don't show the warning until the VM sizes are loaded
+      return '';
+    }
+    const selectedVmIsAvailableInSelectedRegion = this.vmSizes.filter((vmData) => {
+      return vmData.Name === get(this, 'config.size');
```
mantis-toboggan-md, line 353: This looks like the same thing as `selectedVmSizeExistsInSelectedRegion`

lib/nodes/addon/components/driver-azure/component.js
```diff
@@ -139,12 +171,118 @@ export default Component.extend(NodeDriver, {
...
+      .concat({
+        kind:     'group',
+        name:     withAnLabel,
+        value:    withAnLabel,
+        disabled: true
+      })
+      .concat(vmsWithAn)
+
+    if (!get(this, 'selectedVmSizeExistsInSelectedRegion')) {
+      out.push({
```
mantis-toboggan-md, line 265: Unfortunately I don't see _why_, but this doesn't seem to work. When I select a VM size then switch to a region in which the size is not available, I see the warning about the size not being supported, but the dropdown value switches to whatever is first in `sizeChoices` and the invalid size doesn't seem to be in the list (unless I overlooked it, it's a big list)...then when I switch back to the previous region, the dropdown value also switches back to what I had selected.

lib/nodes/addon/components/driver-azure/template.hbs
```diff
@@ -235,6 +331,22 @@
...
         </p>
       </div>
+      <div class="col span-6">
+        <label class="acc-label">
+          {{t "nodeDriver.azure.acceleratedNetworking.label"}}
+        </label>
+        <div>
+          {{input
+            type="checkbox"
+            checked=config.acceleratedNetworking
```
mantis-toboggan-md, line 341: It would be cool if this worked like the rke2 accelerated networking checkbox: disabled when the selected vm size doesn't support accelerated networking and config.acceleratedNetworking is `false`. That can be achieved by adding `disabled=(and (not config.acceleratedNetworking) (not selectedVmSizeSupportsAN))`

lib/nodes/addon/components/driver-azure/template.hbs
```diff
@@ -264,13 +376,35 @@
...
-          classNames="form-control"
-          content=sizeChoices
-          optionLabelPath="value"
-          optionGroupPath="group"
-          value=config.size
-        }}
+        {{#if vmSizesLoading}}
+          <i class="icon icon-spinner icon-spin"></i>
+        {{else}}
+          {{new-select
```
mantis-toboggan-md, line 382: This can be a really long list, it miiiight be better to use a searchable-select; however, that component is a different height than the other input in this row so it is kinda ugly:
Screen Shot 2022-12-19 at 2 37 46 PM

new-select and searchable-select have different padding and afaict that was accidental. seachable-select was given `padding: 10px` to match new-select back in Jan, then in July new-select was switched from `padding: 10px` to `padding: 4px 10px` to fix a clipping issue when the height was forced to match other inputs (text inputs have `padding: 5px 10px`) those PRs are, respectively, https://github.com/rancher/ui/pull/4805 and https://github.com/rancher/ui/pull/4855

I think the fix would be to change searchable-select to `padding: 4px 10px`. But I don't see any of this as essential to this PR so feel free to ignore the entirety of this comment if you'd prefer not to mess with such a widely-used component.

mantis-toboggan-md (changes requested):

lib/nodes/addon/components/driver-azure/template.hbs
catherineluse: removed the vmSizeAvailabilityWarning

lib/nodes/addon/components/driver-azure/component.js
catherineluse: I agree - replaced with:

```js
if (!get(this, 'selectedVmSizeExistsInSelectedRegion')) {
      return intl.t('nodeDriver.azure.size.availabilityWarning');
    }
```

lib/nodes/addon/components/driver-azure/component.js
catherineluse: I refactored it and it seems fixed: 

```js
let out = [{
      kind:     'group',
      name:     noAnLabel,
      value:    noAnLabel,
      disabled: true
    }]
      .concat(vmsWithoutAn)
      .concat({
        kind:     'group',
        name:     withAnLabel,
        value:    withAnLabel,
        disabled: true
      })
      .concat(vmsWithAn)

    out = out.map((vmData) => {
      const { Name } = vmData;

      if (vmData.kind === 'group') {
        return vmData;
      }

      return {
        label:    Name,
        value:    Name,
        disabled: vmData.disabled || false,
      };
    });

    if (!get(this, 'selectedVmSizeExistsInSelectedRegion')) {
      return out.concat({
        label:    get(this, 'config.size'),
        value:    get(this, 'config.size'),
        disabled: true
      });
    }

    return out
```
Maybe later this week I'll take some time to figure out why the refactor works but not the original.

catherineluse: Oh I think I understand. About array.push:

> It changes the size of the original array and returns a new array (with the new element added) as output.

The way I had it, I conditionally added something to the array, then returned the original array. But when I conditionally added the array item, it created a new array. Whereas the original variable was pointing to the outdated array - the one that didn't have the new item added.

lib/nodes/addon/components/driver-azure/template.hbs
catherineluse: Added the disabled prop and a tooltip to make sure it's clear why. 'Accelerated networking is only supported for certain VM sizes.'
Screenshot 2022-12-19 at 3 53 17 PM

lib/nodes/addon/components/driver-azure/template.hbs
catherineluse: I agree that searchable select would be more friendly, but in my setup it's not just the padding that looks weird, but also the height. The searchable select is so tall that it extends past the form:
Screenshot 2022-12-19 at 3 57 41 PM

For fear of making breaking changes to the searchable select, I'll leave it as new-select for now, but I'll open a separate issue about fixing the styling for that component.

mantis-toboggan-md: Oof, yeah, I agree about that being too much to tack onto this PR.

lib/nodes/addon/components/driver-azure/component.js
mantis-toboggan-md: .push appends to the array and returns the new length of the array https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

concat otoh does return a new array https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

Maybe we're missing something about how ember computed properties work? It's weird that the contents of the dropdown are updated when a new region is selected, and just that option isn't added.

lib/nodes/addon/components/driver-azure/template.hbs
mantis-toboggan-md: The one that needs to go is actually `vmAvailabilityZoneWarning` completely my fault, sorry. Other than that the PR looks good

mantis-toboggan-md (changes requested):
I gave a bad suggestion initially - vmAvailabilityZoneWarning should be removed instead of the size warning. Other than that this all looks good.

lib/nodes/addon/components/driver-azure/component.js
catherineluse: I'm pretty sure it's a Javascript thing. What I'm saying is that `const out` was pointing to the memory address of the old array which never got the new item.

catherineluse:
I've updated it to replace `vmAvailabilityZoneWarning` with `showVmSizeAvailabilityWarning`

mantis-toboggan-md (approved):

PR #4938, opened 2023-01-05
Fix confirmation dialog for multiple deactivation

nwmac:
Addresses https://github.com/rancher/dashboard/issues/7762

The 'deactivate' dialog and the way it is wired into the drivers does not support multiple resources - we dispatch the confirmation dialog for each resource individually, so if you have 2 drivers selected, we dispatch twice - the 2nd time closes the dialog from the first. Thus, if you select 3 drivers, you will see the confirmation, but only that 3rd driver will be deactivated if you proceed. 

This PR updates the dialog to show multiple driver names and to deactivate multiple when invoked as such.

It then updates the two driver lists to override the bulk action handler so that we can open the dialog once and pass it multiple items. This is a pattern already used elsewhere in the code.

Finally, I corrected a typo in the action name: `promotDeactivate`.

Example with multiple selected:

image

### DIFF
--- a/lib/nodes/addon/custom-drivers/cluster-drivers/controller.js
+++ b/lib/nodes/addon/custom-drivers/cluster-drivers/controller.js
@@ -60,7 +60,7 @@ export default Controller.extend({
           originalModel: models[0],
           batchModels:   models,
           action:        'deactivate'
-        });        
+        });
       }
     };
   }),
--- a/lib/nodes/addon/custom-drivers/cluster-drivers/controller.js
+++ b/lib/nodes/addon/custom-drivers/cluster-drivers/controller.js
@@ -53,6 +53,18 @@ export default Controller.extend({
     },
   },
 
+  bulkActionHandler: computed(function() {
+    return {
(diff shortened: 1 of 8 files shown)


lib/nodes/addon/custom-drivers/cluster-drivers/controller.js
```diff
@@ -53,6 +53,18 @@ export default Controller.extend({
     },
   },
 
+  bulkActionHandler: computed(function() {
+    return {
+      promptDeactivate: (models) => {
+        get(this, 'modalService').toggleModal('modal-confirm-deactivate', {
+          originalModel: models[0],
+          batchModels:   models,
```
codyrancher, line 61: I don't think it's worth having you chance since it's the old UI. But this seems like an odd interface. When batchModels is used it doesn't look like originalModel is used (did I miss it?). And we could've just maintained the old interface and  differentiated between array of models and single model within the component.

codyrancher (approved):

mantis-toboggan-md (approved):

PR #4942, opened 2023-01-12
Fix imported AKS empty banner

nwmac:
Addresses https://github.com/rancher/dashboard/issues/7781

This PR fixes the above issue.

Simple fix - we are calling the callback with no parameters in the success case, where the callback is expecting a first argument to indicate success - thus undefined is passed in which is treated as an error. The 2nd error argument is also not specified, so we end up pushing `undefined` to the list of errors which results in the empty banner.


mantis-toboggan-md (approved):
This PR fixes the linked issue but I found another problem when I entered bad credentials. It appears unrelated to this change so I filed a separate issue here: https://github.com/rancher/dashboard/issues/7913

PR #4946, opened 2023-01-15
RKE1 pod security admission support

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/7575

This PR adds PSA support during RKE1 configuration. This is only available for  kubernetes versions >=1.23.0. Users creating clusters with versions <1.25 should have the same experience as before, with the addition of a banner informing them that psp support will be disabled in later versions. 

### Creating a >=1.25.0 cluster
* no psp options and a banner explaining why
* psa dropdown with a default of 'privileged'
* field being set is `rancherKubernetesEngineConfig.services.kubeApi.podSecurityConfiguration`

### Upgrading to >=1.25.0 from a version with PSP support
* if psp support is enabled, there should be a banner warning to disable it and an error on attempting to create
* if there are psp resources within the cluster (one or more projects with a psp set) there should be a banner asking the user to delete these psp resources
* psa dropdown as with creating a >=1.23.0 cluster

### rke1 templates
* users should be able to configure the psa config template field

### testing
When I run the ember UI locally websockets do not work correctly: this makes testing cluster provisioning difficult (the ui waits for a message saying the cluster resource is created before creating node pools; this wont work without websockt support). I got around this by building the ember UI and running rancher-in-docker with that custom-built UI replacing the one bundled in the rancher/rancher image using the --mount command. If you don't want to go throught hat ping on slack and I'll give you login details for a testing setup.

### DIFF
--- a/app/authenticated/cluster/edit/route.js
+++ b/app/authenticated/cluster/edit/route.js
@@ -19,6 +19,22 @@ export default Route.extend({
 
     const cluster     = this.modelFor('authenticated.cluster');
 
+    const projects = cluster.projects
+
+    const hasPSPResources = async() => {
+      for (const p of projects){
+        const bindings = await p.followLink('podSecurityPolicyTemplateProjectBindings')
+
+        if (bindings.length){
+          return true
+        }
+      }
+
+      return false
+    }
+
+
+
     let modelOut      = {
       originalCluster:            cluster,
       cluster:                    cluster.clone(),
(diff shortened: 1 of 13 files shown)


codyrancher (dismissed):

nwmac (changes requested):
One change - we want to show the PSA Template drop down if the k8s version is >= 1.23

codyrancher (approved):

nwmac (approved):

PR #4949, opened 2023-01-19
Disable "import host" for local cluster

nwmac:
Addresses https://github.com/rancher/dashboard/issues/4964

Rancher does not manage the local cluster, so the 'Import Host' option should not be shown - currently it is:

image

With this PR it is now not shown:

image

This PR adds an additional check for the local (internal) cluster.


codyrancher (dismissed):

nwmac:
Fixed lint issue - need a re-approval

codyrancher (approved):

PR #4950, opened 2023-01-19
Fix labels not being saved on hosted providers

nwmac:
Addresses https://github.com/rancher/dashboard/issues/7631

This PR fixes the above issue.

We are not sending the labels or annotations for hosted providers.

This PR adds an extra arg to the save method, so that we can determine if the labels or annotations have changed, and only send those it they have.

Tested with an Azure AKS cluster.

### DIFF
--- a/app/models/cluster.js
+++ b/app/models/cluster.js
@@ -1075,6 +1075,21 @@ export default Resource.extend(Grafana, ResourceUsage, {
     return true;
   },
 
+  compareStringArrays(a, b) {
+    let aStr = '';
+    let bStr = '';
+
+    Object.keys(a || {}).forEach((key) => {
+      aStr += `${ key }=${ a[key] },`;
+    });
+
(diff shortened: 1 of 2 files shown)


app/models/cluster.js
```diff
@@ -1075,7 +1075,22 @@ export default Resource.extend(Grafana, ResourceUsage, {
...
+    let aStr = '';
+    let bStr = '';
+
+    Object.keys(a || {}).forEach((key) => {
+      aStr += `${ key }=${ a[key] },`;
+    });
+
+    Object.keys(b || {}).forEach((key) => {
+      bStr += `${ key }=${ b[key] },`;
+    });
```
codyrancher, line 1088: Doesn't this have to be sorted? I didn't think objects in js guarantee order.

nwmac: Good spot - fixed

codyrancher (approved):

PR #4964, opened 2023-02-13
AKS updates: maxSurge, node pool labels, node pool taints, system pools

nwmac:
Addresses: https://github.com/rancher/dashboard/issues/8073
Addresses: https://github.com/rancher/dashboard/issues/8105

For AKS provisioned clusters - this PR:

- Adds support for maxSurge on Node Pools
- Adds support for Labels on Node Pools
- Adds support for Taints on Node Pools
- Adds support for adding system Node Pools
- Removes 'Advanced' section in Networking, so that 'Enable Private Cluster' is more discoverable
- Fixes a bug where not entering a node pool name left the UI stuck in "Saving..."

Notes:
- Taints are persisted as an array of strings in the format: `key=value:effect` - we parse this and format them this way on change
- Lablels and Taints sections will auto-expand on load if labels or taints are set

**To Test**

- Add a cloud credential for Azure
- Go to `Cluster Management`-> `Clusters` -> `Create`. Click `Azure AKS` under "Create a cluster in a hosted Kubernetes provider"
- Enter a cluster name, select your Azure cloud credential and click `Next: Configure Cluster`.
- Choose a Location and click `Configure Cluster`
- Enter a name for the `Cluster Resource Group` and `DNS Prefix`
- Scroll down to `Node Pools`
- You should see `Max Surge`, `Labels` and `Taints` are now available, e.g.

image

- Enter "20%" for Max Surge
- Add two labels: key: `label1`, value `value1` and key: `label2`, value `value2`
- Add a taint: key: `taint1`, value `taintvalue1`, effect: `PreferNoSchedle`

Should look like this:

image

- Click 'Add Node Pool`
- Enter the name `systempool`, set the Mode to `System` and select just Availability Zone 1.

- Click 'Add Node Pool`
- Enter the name `userpool`, set the Mode to `User` and select just Availability Zone 1.
- Enter "15%" for Max Surge
- Add a label: key: `userlabel1`, value `uservalue1`
- Add a taint: key: `usertaint1`, value `usertaintvalue1`, effect: `NoExecute`

- Create the cluster by clicking `Create`
- Wait for the cluster to  finish provisioning

- Open the Azure portal UI go to "Kubernetes Services" and locate the cluster you created above.
- Validate that the node pool settings made above are reflected in the Azure UI
- example of how node pools list should look:
image

Go through each Node Pool checking its Mode, Max Surge, Labels and Taints.

agentpool
image

systempool
image

userpool
image

- Back in Rancher in the Clusters list, `Edit Config` for the cluster created above
- Click `Next: Configure Cluster`
- Validate that the node pool data is shown correctly
- Validate that for the node pools with taints and labels, these sections are shown expanded
- Remove a label and taint from one of the pool. Change the Max Surge.
- Click Save
- Validate that the changes made are reflected when going back into edit in Rancher
- Validate that the changes made are reflected in the Azure Portal UI
- Delete the cluster


lib/shared/addon/components/aks-node-pool-row/component.js
```diff
@@ -27,9 +27,62 @@ export default Component.extend({
...
+        const splitLabel = (splitEffect[0] || '').split('=');
+
+        return {
+          effect: splitEffect[1],
+          key:    splitLabel[0],
+          value:  splitLabel[1],
+        };
+      })
+
+      set(this, 'taints.nodeTaints', _taints);
```
mantis-toboggan-md, line 53: I think I found a solution for the taint bug we saw [here](https://github.com/emberjs/data/issues/3276#issuecomment-110435315)
... it looks like when `nodeTaints` is set this way, the aks-node-pool-row component prototype is mutated, so each instance of aks-node-pool-row is essentially sharing the same array of taints. Going off the solution I linked above, you can fix the taint bug by deleting the `taints` definition on line 31 doing something like 
```
this.taints = {
   nodeTaints: _taints
}
```

and likewise replacing line 56 with 
```
this.taints = {
   nodeTaints: []
}
```

I tried this out myself and creation/editing all seemed fine with various configurations of multiple node pools with and without taints - I'd push the change up myself but don't have permission to do so.

mantis-toboggan-md (changes requested):

lib/shared/addon/components/aks-node-pool-row/component.js
nwmac: Thanks @mantis-toboggan-md - that worked. All updated now and should be ready for review.

mantis-toboggan-md (approved):

PR #4962, opened 2023-02-09
[2.6] azuread migration warning

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/7330#issuecomment-1424855747

This PR adds a warning banner when azure ad is enabled with the old (soon to be deprecated) endpoint configuration, taken from this 2.5 PR: https://github.com/rancher/ui/pull/4873/files The changes to ember azure ad auth configuration are not included here; it needs to be configured through the dash in 2.6.11

Screen Shot 2023-02-09 at 3 06 37 PM

PR #4978, opened 2023-03-01
Add missing constant

nwmac:
Addresses https://github.com/rancher/dashboard/issues/8299

This PR was added in 2.6.11 to add the warning banner: https://github.com/rancher/ui/pull/4962/files

If references the constant `AZURE_AD.ANNOTATION_MIGRATED` - but that is not defined in `constants.js`, so the banner is always shown when azuerad is enabled.

This PR fixes this by adding the missing constant:

```
ANNOTATION_MIGRATED: 'auth.cattle.io/azuread-endpoint-migrated'
```

Validated with a test AzureAD auth provider.

### DIFF
--- a/lib/shared/addon/utils/constants.js
+++ b/lib/shared/addon/utils/constants.js
@@ -753,7 +753,7 @@ C.AZURE_AD = {
     ENDPOINT:       'https://login.chinacloudapi.cn/',
     GRAPH_ENDPOINT: 'https://graph.chinacloudapi.cn/'
   },
-  CUSTOM: { KEY: 'custom' },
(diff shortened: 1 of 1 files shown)


nwmac:
FYI @mantis-toboggan-md - I don't see this in `master` - so assume its not needed for 2.7?

richard-cox (approved):

mantis-toboggan-md:
Correct @nwmac  - the dashboard has this banner in 2.7, and unlike 2.6, it's not possible to bypass the dash and go directly to the ember ui, missing the warning in the process

PR #5009, opened 2023-04-19
Add new available EKS k8s versions to hardcoded list

furkatgofurov7:
Signed-off by: Furkat Gofurov (furkat.gofurov@suse.com)

Updates hardcoded EKS supported k8s versions list with newer versions based on https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html

Since tech debt https://github.com/rancher/dashboard/issues/7295 is still open, we need to manually update this. 

@mjura @pgonin

### DIFF
--- a/lib/shared/addon/utils/amazon.js
+++ b/lib/shared/addon/utils/amazon.js
@@ -683,7 +683,7 @@ export const EKS_REGIONS = [
 ];
 
 // from https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
-export const EKS_VERSIONS = ['1.24', '1.23']; // sort newest->oldest so we dont have to run any logic to sort like other provider versions
+export const EKS_VERSIONS = ['1.26', '1.25', '1.24', '1.23']; // sort newest->oldest so we dont have to run any logic to sort like other provider versions
 
 export const nameFromResource = function(r, idField) {
   let id = r[idField];


alexander-demicev (approved):

richard-cox (changes requested):
Was just about to approve this after bringing up a 1.25 cluster, but noticed a bug in the way we handle the initial value - https://github.com/rancher/dashboard/issues/8721.

Will look for someone our side to fix, but happy for anyone in this PR to check it out (just let us know.

Until that's fixed we can't merge this issue. I'll mark it with `Request changes` just so it's clear we shouldn't merge yet.

@mjura once this is all sorted, did you want to do independent validation before this merges?

Edit - Alternatively, if `ui-k8s-default-version-range` is unlikely to change until we release again, 1.26 could be removed as an option

Edit 2 - FWIW both 1.25 and 1.26 provisioned ok and i could browse resources in the Explorer. Didn't do any thorough tests though

furkatgofurov7:
**DO NOT MERGE, until https://github.com/rancher/ui/pull/5009#pullrequestreview-1396216403 is resolved!**

/hold

Thanks @richard-cox for bringing this up, let's make it clear so that it will not be merged accidentally.

richard-cox (approved):
Removing stale review

PR #5019, opened 2023-05-11
AKS MaxSurge: Address QA feedback

nwmac:
This PR addresses feedback for https://github.com/rancher/dashboard/issues/8073.

It:

- Allows maxSurge to be editable after cluster creation
- Sets defaulr for maxSurge to 1
- Adds validation for maxSurge so it is either a percentage (1-100) or an integer greater than 1


mantis-toboggan-md (approved):
lgtm

PR #5021, opened 2023-05-11
Correct agent config matchExpressions values

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/8856

The `value` field wasn't populating with the default affinity values correctly because the key was wrong, should be `values` not `value`: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#nodeselectorrequirement-v1-core

I've also updated the label in the UI to more accurately reflect the spec: 
Screen Shot 2023-05-11 at 3 35 45 PM


rak-phillip (approved):
LGTM

PR #5022, opened 2023-05-15
fix agent config cpu req unit formatting

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/8871
https://github.com/rancher/dashboard/issues/8864

This PR updates the unit parsing/formatting for cpu requests to use the same utility function used for other mCPU inputs around the Ember UI. The previous solution implemented runs into problems w/ floating point arithmetic accuracy, eg when the value inputted is `1001m` this is parsed to `1.001` and `1.001*1000` returns `1000.99...`. 

Screen Shot 2023-05-15 at 10 59 54 AM


rak-phillip (approved):
LGTM

PR #5029, opened 2023-05-24
Allow legacy login page to be used by Rancher CLI

nwmac:
Addresses https://github.com/rancher/dashboard/issues/8964

Re-enabled the legacy login page if the URL contains a query string - this ensures it can be used by the Rancher CLI.


mantis-toboggan-md (approved):
lgtm - tested with steps in the issue: https://github.com/rancher/dashboard/issues/8964#issuecomment-1561131738

PR #5031, opened 2023-05-26
fix eks edit removal of agent config fields

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/8968#issuecomment-1560366808

When EKS clusters are edited, only changed values are sent in the request and the rest are preserved (more info on why [here](https://github.com/rancher/rancher/issues/28541)). When a user clears out previously-defined cluster/fleet agent customization the relevant fields are deleted from the cluster spec; consequently, with imported EKS clusters specifically, the previously-set values are preserved. 

This PR updates the EKS cluster save function to set any cluster/fleet agent fields the user has cleared out to `null`, which will properly overwrite the previous value.

I found another bug in the same area (editing agent config) that appears to have a different underlying cause, issue here: https://github.com/rancher/dashboard/issues/9003

#### Testing

1. Edit an imported EKS cluster to add cluster/fleet agent config values
2. Edit again and remove values
3. Open the edit view and verify that the values were removed as expected


codyrancher:
I understand the change and I'm okay with it going in. I had one question though. Does the backend return the null values?

codyrancher (approved):

PR #5044, opened 2023-06-27
Fix deprecation warnings

nwmac:
WIP

Updates lint rule so string prototype extensions fail the lint.

Addresses string prototype extension uses and uses `@ember\string` functions.

Note: This PR is based off of the credit card dependency removal PR - this should be merged first.


app/models/nodetemplate.js
```diff
@@ -125,7 +125,7 @@ export default Resource.extend({
 
       this.registerDynamicComputedProperty('displaySize', computedKeys, size.getDisplayProperty);
     } else {
-      set(this, 'displaySize', get(this, 'config.size') || 'N/A');
+      // set(this, 'displaySize', get(this, 'config.size') || 'N/A');
```
mantis-toboggan-md, line 128: Was this included intentionally? Seems unrelated

lib/pipeline/addon/components/pipeline-codemirror/component.js
mantis-toboggan-md, line 1: This component is still referenced in another component. Pipelines have been deprecated since 2.5, can we remove all pipeline code now?? Probably deserves its own PR though

mantis-toboggan-md (changes requested):

app/models/nodetemplate.js
nwmac: No, that was a test - fixed

lib/pipeline/addon/components/pipeline-codemirror/component.js
nwmac: Yeah - we should do that in a separate PR

nwmac: I'll put it back in

PR #5046, opened 2023-06-30
Add Data-testid attributes to support agent config e2e tests

mantis-toboggan-md:
Needed for https://github.com/rancher/dashboard/pull/9247

### DIFF
--- a/lib/shared/addon/components/form-match-expressions-k8s/component.js
+++ b/lib/shared/addon/components/form-match-expressions-k8s/component.js
@@ -99,9 +99,13 @@ export default Component.extend({
 
     operatorChanged(match, selected){
       if (selected.value === 'Exists' || selected.value === 'DoesNotExist'){
+        // set values to empty array to force input component to update
+        set(match, 'match.values', [])
+        delete match.match.values
+      } else if (!match.match.values) {
         match.match.values = []
       }
-    }
+    },
   },
 
   // node match expressions/fields have a few more operator options
--- a/lib/shared/addon/components/form-match-expressions-k8s/component.js
+++ b/lib/shared/addon/components/form-match-expressions-k8s/component.js
@@ -13,6 +13,8 @@ const MATCH_TYPES = {
 
 export default Component.extend({
   layout,
+  attributeBindings: ['data-testid'],
+
   editing:          false,
   matchExpressions: null, // node selector term
   matchFields:      null, // node selector term
--- a/lib/shared/addon/components/form-node-selector-term-k8s/component.js
+++ b/lib/shared/addon/components/form-node-selector-term-k8s/component.js
@@ -79,8 +79,7 @@ export default Component.extend({
     },
     set(key, val){
       if ((!val || !val.length) && this.nodeSelectorTerm.matchExpressions){
-        set(this, 'nodeSelectorTerm.matchExpressions', null)
-
+        delete this.nodeSelectorTerm.matchExpressions
         this.notifyPropertyChange('nodeSelectorTerm')
       } else if (val && val.length){
         set(this.nodeSelectorTerm, 'matchExpressions', val)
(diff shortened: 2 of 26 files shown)


richard-cox (changes requested):
The e2e relates changes look fine, however there's some functional changes to the files below. Are they part of another PR?
- lib/shared/addon/components/form-match-expressions-k8s/component.js
- lib/shared/addon/components/form-match-expressions-k8s/template.hbs
- lib/shared/addon/components/form-pod-affinity-k8s/component.js
- lib/shared/addon/components/form-pod-affinity-term-k8s/component.js
- lib/shared/addon/components/input-array-as-string/component.js

lib/shared/addon/components/form-pod-affinity-term-k8s/component.js
```diff
@@ -189,12 +186,12 @@ export default Component.extend({
       switch (val){
       case namespaceModes.ALL:
         set(this, 'namespaceSelector', {})
-        set(this, 'namespaces', null)
+        delete this.podAffintyTerm.namespaces
```
mantis-toboggan-md, line 189: Setting to null or deleting the `namespaces` key should have the same effect - I've updated this to delete the key to match the rke2 behavior.

lib/shared/addon/components/form-pod-affinity-k8s/component.js
```diff
@@ -59,7 +60,10 @@ export default Component.extend({
       const neu = {
         priority: TERM_PRIORITY.REQUIRED,
         anti:     false,
-        term:     { _id: randomStr() }
+        term:     {
+          _id:        randomStr(),
+          namespaces: null
```
mantis-toboggan-md, line 65: Added the `namespaces` field here to match the rke2 behavior

lib/shared/addon/components/form-match-expressions-k8s/component.js
```diff
@@ -97,9 +99,13 @@ export default Component.extend({
 
     operatorChanged(match, selected){
       if (selected.value === 'Exists' || selected.value === 'DoesNotExist'){
+        // set values to empty array to force input component to update
+        set(match, 'match.values', [])
+        delete match.match.values
+      } else if (!match.match.values) {
```
mantis-toboggan-md, line 105: If I've read the spec correctly an empty array is acceptable here but I updated this to remove the key instead to match rke2 behavior. This caused a bug in which setting the operator to `In List`, defining `values`,  switching to `Exists` or `DoesNotExist`, then back to `In List` would remove `values` from the match expression, but the input would still show the previously entered text - setting an empty array first forces the input to clear itself and better reflect the actual value of `values`

lib/shared/addon/components/form-node-selector-term-k8s/component.js
```diff
@@ -78,8 +79,7 @@ export default Component.extend({
     },
     set(key, val){
       if ((!val || !val.length) && this.nodeSelectorTerm.matchExpressions){
-        set(this, 'nodeSelectorTerm.matchExpressions', null)
-
+        delete this.nodeSelectorTerm.matchExpressions
```
mantis-toboggan-md, line 82: Technically setting these fields to `null` should work the same as deleting the keys but results in a payload shaped differently than it would in rke2

mantis-toboggan-md:
I've rebased off recent other work and added some comments explaining changes other than the data-testid attributes. I made some tweaks to the payload sent to better align with the rke2 side of things because I thought it'd be nicer UX if the two forms that look the same behaved exactly the same under the hood. I can revert these changes (and add a second set of payload comparison data on the dashboard side) if you think it's messy to include as part of testing PRs

lib/shared/addon/components/form-match-expressions-k8s/component.js
aalves08: Agree with the change

lib/shared/addon/components/form-node-selector-term-k8s/component.js
aalves08: agree

lib/shared/addon/components/form-pod-affinity-k8s/component.js
aalves08: agree

lib/shared/addon/components/form-pod-affinity-term-k8s/component.js
aalves08: agree

aalves08 (approved):
All changes to match rke2 work looks good. Overall, lgtm. 👍

PR #5048, opened 2023-07-13
Remove unused dependencies

mantis-toboggan-md:
part of https://github.com/rancher/dashboard/issues/8067 - this PR removes a few dependencies no longer used. 
card - credit card form
cytoscape - istio v1 visuals
forever-agent - old implementation of server/proxies/api.js
liquid-fire - animated transitions (no longer present)
mysql, sendgrid - caas api user registration components no longer present (https://github.com/rancher/ui/pull/1023)


richard-cox (approved):
Tested the core embedded functionality
- Cluster Manager features, specifically creating an RKE1 cluster and node template
- `Legacy` cluster context features
- `Legacy Apps` features
Also that the old ui is reachable (even though we don't support it anymore)

I've also tried to go back in time to check when the specific changes came in, some were quite hard to track down.

PR #5049, opened 2023-07-14
add aks node resource group input field and length validation

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/8980

This PR adds an input to the AKS provisioning form for the new `nodeResourceGroup` field.  You can read more about what this field is here https://github.com/rancher/aks-operator/pull/210. Per the backend issue, this field is only editable on create (available "when creating a new aks cluster"); this is in line with the cluster resource group name.

To test the new field:
1. Begin provisioning a new AKS cluster: required fields are name, resource group and dns prefix - use something like `<initials>-resource-group` and `<initials>-dns-prefix`
2. Add a node resource group value that exceeds 80 characters
3. Verify that the cluster does not save and an error is shown
4. Reduce the length of the node resource group 
5. Save and verify that the POST request payload and response both include `aksConfig.nodeResourceGroup`
6. Once the cluster begins provisioning, navigate to https://portal.azure.com/#view/HubsExtension/BrowseResourceGroups and observe that both the cluster and node resource group names you configured have been created
7. Edit the cluster and verify that the field is displayed, but uneditable

To test default behavior, verify that an AKS cluster will provision without configuring the node resource group field. 

Screen Shot 2023-07-14 at 12 12 02 PM
Screen Shot 2023-07-14 at 12 12 10 PM
Screen Shot 2023-07-14 at 12 12 49 PM

view mode:
Screen Shot 2023-07-14 at 12 11 05 PM

### DIFF
--- a/lib/shared/addon/components/cluster-driver/driver-aks/component.js
+++ b/lib/shared/addon/components/cluster-driver/driver-aks/component.js
@@ -610,15 +610,17 @@ export default Component.extend(ClusterDriver, {
     let model = get(this, 'cluster');
     let clusterErrors = model.validationErrors() || [];
     const nodePoolErrors = this.validateNodePools();
-
     const vnetSet = !!get(this, 'config.virtualNetwork');
-
     if (vnetSet) {
       clusterErrors = clusterErrors.concat(this.validateVnetInputs());
     }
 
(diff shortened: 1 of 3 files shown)


momesgin (approved):
LGTM

PR #5052, opened 2023-07-25
Bugfix agent config

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/9003

In addition to the issue above this PR also fixes an issue where switching an affinity term between from required to preferred clears out other required terms, and some instances in which the match expressions/field select appeared in pod affinity terms (which should only be available in node affinity terms)

### DIFF
--- a/lib/shared/addon/components/form-match-expressions-k8s/template.hbs
+++ b/lib/shared/addon/components/form-match-expressions-k8s/template.hbs
@@ -98,8 +98,7 @@
                 value=match.match.values
               }}
             {{else}}
-            asdasd
-            {{t "generic.na"}}
+              {{t "generic.na"}}
             {{/if}}
           {{/input-or-display}}
         </td>
--- a/lib/shared/addon/components/form-match-expressions-k8s/template.hbs
+++ b/lib/shared/addon/components/form-match-expressions-k8s/template.hbs
@@ -1,7 +1,7 @@
 <table class="table fixed no-lines mb-20">
   <thead>
     <tr>
-      {{#if (not labelSelector)}}
+      {{#if (not isPod)}}
         <th class="acc-label">
           {{t "clusterNew.agentConfig.overrideAffinity.nodeAffinity.nodeSelectorTerm.type"}}
         </th>
@@ -25,7 +25,7 @@
   <tbody>
     {{#each allMatches as |match|}}
       <tr>
-        {{#if (not labelSelector)}}
+        {{#if (not isPod)}}
           <td>
           {{#input-or-display
               editable=editing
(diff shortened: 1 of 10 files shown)


aalves08 (dismissed):
Gave it a test and bug looks fixed. Codewise lgtm!

lib/shared/addon/components/form-match-expressions-k8s/template.hbs
```diff
@@ -98,6 +98,7 @@
                 value=match.match.values
               }}
             {{else}}
+            asdasd
```
rak-phillip, line 101: It looks like this needs to be removed

rak-phillip (changes requested):
I only see one thing that needs to be cleaned up, otherwise the changes look good to me. 

I'm not sure if I quite understand the reasoning behind replacing `labelSelector` with `isPod` throughout this change. My assumption at this point is that `labelSelector` was being used incorrectly throughout this point. Is that a correct assessment of this fix?

lib/shared/addon/components/form-match-expressions-k8s/template.hbs
aalves08: Good catch @rak-phillip

mantis-toboggan-md: oof, thanks, it's gone now

lib/shared/addon/components/form-pod-affinity-term-k8s/component.js
```diff
@@ -121,7 +121,7 @@ export default Component.extend({
     }
   }),
 
-  labelSelector: computed('podAffintyTerm.labelSelector.matchExpressions', {
+  matchExpressions: computed('podAffintyTerm.labelSelector.matchExpressions', {
```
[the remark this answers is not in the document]
mantis-toboggan-md: In my original match expressions implementation I passed through the whole `labelSelector` object, but this was needlessly complex given that the `matchExpressions` array passed in by the node affinity term component is _almost_ the same as `labelSelector.matchExpressions` (pod and node affinity are full of fields that are _almost_ the same, it's very...fun). I updated this pod affinity component to pass in only the match exp array but I didn't properly refactor the match expressions component.

mantis-toboggan-md:
>I'm not sure if I quite understand the reasoning behind replacing labelSelector with isPod throughout this change. My assumption at this point is that labelSelector was being used incorrectly throughout this point.

Yes - using the entire labelSelector object in the match expressions component in the pod affinity context and only the match expression array in the node affinity context introduced complexity with no real benefit, but the component still needs to track whether its being used in the pod or node affinity context as the operator options are different, hence, the `isPod` boolean

rak-phillip (approved):
LGTM

Thanks for the clarification on these changes!

PR #5061, opened 2023-08-31
9507 - [GKE] Fixed inconsistency between node and cluster versions

momesgin:
Proposed changes
======
Set the initial node version to the cluster's max valid version, also disabled upgradeVersion when cluster version is higher than the max valid version.

Fixes [#9507](https://github.com/rancher/dashboard/issues/9507)

https://github.com/rancher/ui/assets/135728925/b08fccff-2fa3-49aa-ac80-a11a7ebf3b5f


mantis-toboggan-md (approved):
lgtm

PR #5062, opened 2023-09-01
Show banner that Weave is deprecated for k8s >= 1.27

nwmac:
Addresses: https://github.com/rancher/dashboard/issues/9556

Shows a deprecation banner if the user selects Weave network and a Kubernetes version >= 1.27

This will need manual testing as it requires a configured cloud credential to get to cluster provisioning.

### DIFF
--- a/lib/shared/addon/components/form-network-config/component.js
+++ b/lib/shared/addon/components/form-network-config/component.js
@@ -55,6 +55,7 @@ export default Component.extend({
   enableNetworkPolicy:     null,
   clusterTemplateRevision: null,
   applyClusterTemplate:    null,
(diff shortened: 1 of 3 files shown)


mantis-toboggan-md (dismissed):
Approving anyway because it's nitpicky of me but it looks like this could be contained in a computed prop for better readability:

```
  weaveDeprecated: computed('config.network.plugin', 'config.kubernetesVersion', function(){
    const plugin = get(this, 'config.network.plugin');
    const kubernetesVersion = get(this, 'config.kubernetesVersion');

    return (plugin === WEAVE && gte(coerceVersion(kubernetesVersion), 'v1.27.0'));
  }),
```

PR #5064, opened 2023-09-05
Remove support for aws in-tree provider for k8s >=1.27

mantis-toboggan-md:
RKE1 side of https://github.com/rancher/dashboard/issues/9204

Requirements from internal design docs:

If k8s version >= 1.27 is selected during cluster creation, the aws in-tree provider option should be hidden.
If k8s version >=1.27 is selected while editing a cluster and the aws in-tree provider had previously been configured, the provider options are made editable, and the provder is switched to 'external.' 
If the k8s version is changed back to <1.27, the cloud provider choice is reverted.

Since 1.27 isn't available yet, you can manually add the option [here](https://github.com/mantis-toboggan-md/ui/blob/5551da85705b1101d2cc84dadd13f73cf72e98bd/lib/shared/addon/components/form-versions/component.js#L144)
```
    mappedVersions.push({
      disbaled:     false,
      experimental: false,
      label:        'v1.27.0-rancher1-1',
      value:        'v1.27.0-rancher1-1'
    })
```

https://github.com/rancher/ui/assets/42977925/5167de31-b2d4-473e-a889-3f2119343cb4

https://github.com/rancher/ui/assets/42977925/7c98d848-19cd-40c3-97a9-b94ebf5f3c42

### DIFF
--- a/lib/shared/addon/components/cru-cloud-provider/component.js
+++ b/lib/shared/addon/components/cru-cloud-provider/component.js
@@ -231,6 +231,12 @@ export default Component.extend({
     return !get(this, 'model.harvesterNodeTemplateId') || !get(this, 'cluster.name') || isExternalCredential
   }),
 
+  awsSupported: computed('cluster.rancherKubernetesEngineConfig.kubernetesVersion', function(){
+    const kubernetesVersion = get(this, 'cluster.rancherKubernetesEngineConfig.kubernetesVersion')
+
+    return Semver.lt(Semver.coerce(kubernetesVersion), '1.27.0')
+  }),
+
   checkDefaults(record) {
     get(this, 'azureDefaults').forEach((def) => {
       if (isEmpty(record[def])) {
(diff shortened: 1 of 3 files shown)


lib/shared/addon/components/cru-cloud-provider/component.js
```diff
@@ -142,21 +170,10 @@ export default Component.extend({
...
-      } else {
-        next(() => {
-          if (!this.configName) {
-            if (this.isDestroyed || this.isDestroying) {
-              return;
-            }
-
-            set(this, 'selectedCloudProvider', this.selectedCloudProvider === 'generic' ? 'generic' : 'none');
-          }
-        });
```
mantis-toboggan-md, line 159: I couldn't work out what this was attempting to do. It breaks toggling between 'none' and 'external.' I played around with cluster creation/upgrading, with and without templates, couldn't find any problems removing this.

codyrancher: This entire function is difficult to understand.  I'll play around some with the template revision so see if I can tease out the intention

codyrancher (approved):

lib/shared/addon/components/cru-cloud-provider/component.js
codyrancher: I never found a way to run into issues with the removal of code either so I figure it's fine until someone else runs into an issue

PR #5082, opened 2023-11-01
Fix external Amazon banners

nwmac:
Addresses https://github.com/rancher/dashboard/issues/9204

This PR:

- Splits out the banner so that the banner telling you that we changed the cloud provider to 'External Amazon' is seperate
- Shows the banners telling you that you need to configure for external Amazon or external for both create and edit
- Hides the banner that we changed for External Amazon when the user subsequently changes it

To Test:

- Go to create a AWS RKE1 cluster with k8s 1.26 and the 'Amazon (In-Tree)' cloud procider
- On this form, if you change to External Amazon or External, the banner telling you extra config is needed - 

image

image

- Go ahead and create the cluser
- You don't have to wait for create to complete - go to edit the cluster
- Change the k8s version to 1.27,  you should see 2 banners:
image

- Change the cloud provider - the first banner will go away and the appropriate banner will stay for External Amazon and External (same as above)

### DIFF
--- a/lib/shared/addon/components/radio-button/component.js
+++ b/lib/shared/addon/components/radio-button/component.js
@@ -14,6 +14,9 @@ export default Component.extend({
   }),
   click() {
     this.set('selection', this.get('value'));
+
+    this.chosen();
   },
 
+  chosen() {}
 });
--- a/lib/shared/addon/components/cru-cloud-provider/component.js
+++ b/lib/shared/addon/components/cru-cloud-provider/component.js
@@ -35,7 +35,8 @@ export default Component.extend({
   azureDefaults,
   azureDescriptions:           AzureInfo,
   unsupportedProviderSelected: false,
-  // track the original configuration to revert the switch to 'external' when the selected provider isnt supported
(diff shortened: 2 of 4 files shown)


lib/shared/addon/components/radio-button/component.js
```diff
@@ -14,6 +14,9 @@ export default Component.extend({
   }),
   click() {
     this.set('selection', this.get('value'));
+
+    this.chosen();
```
codyrancher, line 18: Noop?

codyrancher (approved):
It looks fine. I just wanted to make sure the noop wasn't intended to be something else.

lib/shared/addon/components/radio-button/component.js
nwmac: Yeah - when clicked, it should invoke chosen, which has an empty default which is then over-ridden by uses of the component.

codyrancher: lgtm

PR #5092, opened 2023-12-07
Fix GKE Windows node pool image

nwmac:
Addresses https://github.com/rancher/dashboard/issues/10088

See: https://cloud.google.com/kubernetes-engine/docs/deprecations/docker-containerd

This PR:

- Updates the image list for GKE to include the Windows containerd variants
- Adds filtering of the image list to not show the Windows Docker variants based on the k8s version
- Adds logic so that when the user changes the k8s version, the containerd version will be selected if the docker version was previously selected

Requires manual testing as this requires GKE


mantis-toboggan-md (approved):
lgtm

PR #5100, opened 2024-01-16
Removing all references to PSP since they will no longer be supported

codyrancher:
Proposed changes
======
Removes dead code and unnecessary messaging since PSPs will no longer be supported going forward.

Types of changes
======
Breaking change (fix or feature that would cause existing functionality to not work as expected)

Linked Issues
======
https://github.com/rancher/dashboard/issues/9945


mantis-toboggan-md (changes requested):
One minor thing, it looks like [these](https://github.com/rancher/ui/blob/master/lib/shared/addon/utils/constants.js#L356-L364) constants can be removed. Otherwise looks good; I couldn't see any broken pages nor other references to psps

mantis-toboggan-md (approved):

PR #5101, opened 2024-01-17
Fix GKE range name and CidrBlock conflict

nwmac:
Addresses https://github.com/rancher/dashboard/issues/8749

This PR fixes the above bug:

- Ensures that if you use a range name, we do NOT also send the ipv4 cidr block as well
- Ensures that we add a '<None>' option to the range selector, so that a user can clear the range name and set the ipv4 cidr block if they select a range - currently once you select a range name, there is no way to clear it

This requires manual test as you need access to GKE.

Instructions to test are in the issue at the top - you need to create a VPN with subnets that have secondary ranges. Then create a cluster, selecting a secondary range name and validate that the cluster creates okay.

### DIFF
--- a/lib/shared/addon/components/cluster-driver/driver-gke/component.js
+++ b/lib/shared/addon/components/cluster-driver/driver-gke/component.js
@@ -755,13 +755,21 @@ export default Component.extend(ClusterDriver, {
     if (subnetworkMatch) {
       const { secondaryIpRanges = [] } = subnetworkMatch;
 
-      return secondaryIpRanges.map((s) => {
+      const items = secondaryIpRanges.map((s) => {
         return {
           label:       `${ s.rangeName }(${ s.ipCidrRange })`,
           value:       s.rangeName,
           ipCidrRange: s.ipCidrRange,
         }
       });
+
+      items.unshift({
+        label: this.intl.t('clusterNew.googlegke.ipAllocationPolicy.noSecondaryRange'),
+        value: null,
+        ipCidrRange: '',
+      });
+
+      return items;
     }
 
     return [];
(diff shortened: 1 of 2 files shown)


lib/shared/addon/components/cluster-driver/driver-gke/component.js
```diff
@@ -954,6 +962,14 @@ export default Component.extend(ClusterDriver, {
...
     }
 
+    if (get(config, 'ipAllocationPolicy.clusterSecondaryRangeName') && get(config, 'ipAllocationPolicy.clusterIpv4CidrBlock')) {
+      set(config, 'ipAllocationPolicy.clusterIpv4CidrBlock', null);
+    }
+
+    if (get(config, 'ipAllocationPolicy.servicesSecondaryRangeName') && get(config, 'ipAllocationPolicy.servicesIpv4CidrBlock')) {
+      set(config, 'ipAllocationPolicy.servicesIpv4CidrBlock', null);
+    }
+
```
codyrancher, line 972: My networking knowledge is really poor so this might be a reasonable path but it seems odd to me. 

I would think we'd want the UX to either make these fields mutually exclusive or to inform the user that there was a collision via an error message so they could remedy it as they see fit.

Feel free to ignore if my assessment is wrong.

codyrancher (approved):
Verified that we can now deselect the secondary ranges and that we reset the cidr block if the secondary range is present for pod/services secondary range.

image

lib/shared/addon/components/cluster-driver/driver-gke/component.js
nwmac: In the UI, if you select a range, we disable the manual input option and replace it with the IP of the range selected.

I think that's okay - a better UX would probably be radio buttons to explicitly chose, but given Ember is legacy, this is the minimum to fix the isue.

PR #5103, opened 2024-01-17
Add node pool/group duplicate name validation for hosted providers

nwmac:
Addresses https://github.com/rancher/dashboard/issues/9987
Addresses https://github.com/rancher/dashboard/issues/4465

This PR adds node pool/group duplicate name validation for gke, aks and eks.

Same pattern is used in each - for simplicity, added to each provider, rather than attempt to factor into some common code. Ember code is legacy, so this feels like a reasonable approach to address this issue.

Code is straight-forward - adds to the validation method to check each node pool/group name - adds an error with the names of the duplicates if found.

To test, for each provider, add several node pools/groups with some duplicates and validate that you see the error message. Pleas also test with node groups without names.


codyrancher (approved):
Verified in GKE. I don't seem to have access to the other two cloud providers at the moment.

PR #5112, opened 2024-01-18
Update aws regions and add script to automate this

nwmac:
Addresses https://github.com/rancher/dashboard/issues/10294

This PR updates the list of hard-coded AWS regions for EC2 and EKS.

I've added a script which can be run to automatically do this.

Regions are sorted alphabetically across the 3 partitions that we use, so going forward it will be much easier to see the regions added/removed - for this change, you'll see more change as the region list is not currently in any order that is stable (i.e. not alphabetically sorted).

If we wished, we could set up a scheduled GH action to run this script and automatically create a PR is there are changes. We could do this for other data like this - this is at least an improvement ahead of dynamically fetching data (where this is possible).

Note, Ember UI is legacy, so we probably don't want to invest in dynamic lookup and should do that was we continue to port functionality over to dashboard.


codyrancher (approved):
Instead of automating the creation of PRs I really think these lists should be eliminated from the code base entirely. It seems silly that users of our UI could potentially not use new regions just because they don't receive a new release from us.

Ideally we would dynamically retrieve these. If dynamic retrieval isn't an option maybe we should think about storing these in a global setting or something else that can be updated without a new release.

nwmac:
@codyrancher Yeah - as we migrate to Vue these should be dynamic.

PR automation was just a stop-gap suggestion for as long as we have Ember around.

PR #5115, opened 2024-02-15
Show info banner when upgrading Kubernetes 1.27 or greater

rak-phillip:
This displays an info banner that informs users about changes regarding in-tree cloud providers when upgrading to a version of Kubernetes >=1.27 from a version that is <=1.27.

image

closes rancher/dashboard#10414

### DIFF
--- a/lib/shared/addon/components/managed-import-cluster-info/component.js
+++ b/lib/shared/addon/components/managed-import-cluster-info/component.js
@@ -19,7 +19,24 @@ export default Component.extend({
   editing:     false,
   configField: 'k3sConfig',
 
+  initialKubernetesVersion: null,
   supportedK8sVersionRange: alias(`settings.${ C.SETTING.VERSION_K8S_SUPPORTED_RANGE }`),
+
+  init() {
+    this._super(...arguments);
+
+    this.set('initialKubernetesVersion', get(this, `config.kubernetesVersion`))
+  },
+
+  showKubernetesVersionBanner: computed('config.kubernetesVersion', 'initialKubernetesVersion', function() {
+    const initialVersion = get(this, 'initialKubernetesVersion');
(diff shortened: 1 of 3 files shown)


rak-phillip:
I left notes in rancher/dashboard#10414, but it might be worth repeating here. 

Since local clusters are imported, it's not possible for us to get the information that we need to determine where the cluster is running and whether or not it's in-tree or out-of-tree. We decided to compromise by displaying an informational message to notify users of this change in behavior for 1.27+ and allow them to make the decision for themselves if action needs to be taken before moving forward.

lib/shared/addon/components/managed-import-cluster-info/component.js
```diff
@@ -19,8 +19,25 @@ export default Component.extend({
...
+    this._super(...arguments);
+
+    this.set('initialKubernetesVersion', get(this, `config.kubernetesVersion`))
+  },
+
+  showKubernetesVersionBanner: computed('config.kubernetesVersion', 'initialKubernetesVersion', function() {
+    const initialVersion = get(this, 'initialKubernetesVersion');
+    const selectedVersion = get(this, 'config.kubernetesVersion');
+
+    const initialIsGte127 = initialVersion && Semver.gte(initialVersion, '1.27.0');
```
codyrancher, line 35: Wouldn't we want this to be lt(..., '1.27.0') since the banner should show be shown to anyone upgrading to 1.27 or above?

codyrancher (dismissed):

lib/shared/addon/components/managed-import-cluster-info/component.js
rak-phillip: Yeah, that's more straightforward than than using gte and then testing for the inverse below.. should be fixed now

codyrancher (approved):

PR #5129, opened 2024-03-11
Use `initialCloudProvider` for comparison when upgrading Kubernetes from <1.27

rak-phillip:
This uses the `initialCloudProvider` to determine if the cloud provider needs to be set to `external-aws` when changing Kubernetes versions.

The previous implementation that used the selected cloud provider would toggle `selectedCloudProvider` between `external-aws` and `amazonec2` when selecting different kubernetes versions that are >= 1.27.0.

closes rancher/dashboard#10516


mantis-toboggan-md (approved):
LGTM

PR #5132, opened 2024-03-27
Move test and lints to GitHub actions

nwmac:
This PR moves the test and lint from drone to GitHub actions.


richard-cox (approved):
Looks like the old drone ci task is mandatory. I lack admin rights on this repo, so can't force it through, or update Settings --> Branches --> master --> Require status checks to pass before merging

PR #5133, opened 2024-04-08
update drone jobs to run on v2.9.* tags

eva-vashkevich:


rak-phillip (approved):
LGTM - this is routine release maintenance

PR #5151, opened 2024-07-08
Fix GKE default version and tweak node pool upgrade versioning logic

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/11357

There are two semi-related issues in 11357
1. The cluster kubernetes version (sometimes called control plane version in the code) is being set to the latest version google provides (1.29 right now) instead of the latest version Rancher supports (1.28 in 2.8-head)
2. The node pool versions are being set to the latest available version instead of syncing with the control plane version

Fix for problem 1 was fairly straightforward: instead of using the defaultClusterVersion from the GCP API call we set the cluster version to the latest version shown in the version dropdown. In addition to respecting Rancher-specific versioning limitations this has the added bonus of ensuring that the cluster version is _never_ automatically set to something not shown in the UI.

The fix for problem 1 resulted in a few errors in the node pool component because the cluster version was no longer set before the rest of the page (including np components) rendered. Between that and needing to tweak the node pool version selection logic I ended up rewriting a bunch of observers/computed props around the node pool versioning

Expected node pool version behavior:
1. New node pools will have the same version as the cluster. When the cluster version changes, new node pool versions should be updated as well.
2. When the cluster version is updated, existing node pools will have the option of updating to the latest node pool version that is <= the cluster version (this is the restriction present when editing node pools through the google UI)
3. When the node pool upgrade checkbox is _un_checked, the node pool version should revert back

Note about node pool versions:
Technically the gcp api provides a list of node versions and a list of control plane versions. In my own testing I did not find an example of a version present in the control plane list and _not_ the node pool list. Nevertheless, the version offered in the node pool upgrade input is from the node pool list: we're using the highest node pool version that is <= the current cluster (control plane) version. It just happens that this is always equal to the control plane version.

### DIFF
--- a/lib/shared/addon/components/cluster-driver/driver-gke/component.js
+++ b/lib/shared/addon/components/cluster-driver/driver-gke/component.js
@@ -242,9 +242,6 @@ export default Component.extend(ClusterDriver, {
           }
         }
 
-        if (isEmpty(config.kubernetesVersion)) {
-          set(this, 'config.kubernetesVersion', versions?.defaultClusterVersion);
-        }
 
         cb(true);
       }).catch((err) => {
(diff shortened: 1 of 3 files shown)


momesgin (approved):
tested and LGTM

PR #5157, opened 2024-07-30
clean up l18n

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/9361

The assorted "ICU argument missing..." errors shown when running the UI locally are mostly a result of translation strings in en-us being updated to no longer reference arguments still present in other l18n files. This seems to throw errors even if the arguments are still be provided in template files.  Some translation strings have drifted so much in en-us that it was unclear to me how to rewrite the corresponding translations to remove arguments in other files, so I wound up removing them entirely.

Other things going on:

- A bunch of shibboleth translations still referenced in templates were removed from en-us, maybe unintentionally? https://github.com/rancher/ui/pull/3721/files#diff-93bba6cdcf2bf03eb02ffd57e800a775e4b3879683bad0d0f4f3f7508fa92fd7L721

- `modalDrainNode` was renamed to `drainNode` in en-us - this PR updates other l18n files accordingly

### DIFF
--- a/translations/zh-hans.yaml
+++ b/translations/zh-hans.yaml
@@ -4253,7 +4253,6 @@ clusterNew:
       text: 编辑主机选项将更新主机注册命令
     command:
       instructions: '复制以下命令在主机的 SSH 终端运行。'
-      winInstructions: '在一台或者多台运行了受支持的 Docker 版本的主机上运行<code>CMD</code>命令，并且 windows server 版本大于<code class="code-version">{version}</code>.'
     auth:
       label: 认证提供者
       x509: x509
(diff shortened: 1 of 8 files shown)


translations/zh-hans.yaml
```diff
@@ -4253,7 +4253,6 @@ clusterNew:
       text: 编辑主机选项将更新主机注册命令
     command:
       instructions: '复制以下命令在主机的 SSH 终端运行。'
-      winInstructions: '在一台或者多台运行了受支持的 Docker 版本的主机上运行<code>CMD</code>命令，并且 windows server 版本大于<code class="code-version">{version}</code>.'
```
momesgin, line 4256: we have this use case for it:
https://github.com/momesgin/ui/blob/master/lib/shared/addon/components/custom-command/template.hbs#L223
can we still remove it?

mantis-toboggan-md: It should fall back to the translation in the en-us file. Having arguments in other l18n files and not en-us caused some of the console errors in https://github.com/rancher/dashboard/issues/9361. In other cases I was able to update the l18n files to no longer use the argument in question, but with this line, the change in en-us was grammatically too complex for me to guess at how to replicate it.

momesgin (approved):

PR #5160, opened 2024-08-06
Update azure cloud provider option based on k8s version support

momesgin:
Fixes https://github.com/rancher/dashboard/issues/11520

## Occurred changes

- Azure is removed from the list of cloud providers when k8s version is `>= 1.30`
- When selected k8s version is  `>=1.27 or < 1.30` the Azure option will be displayed
- While the supported version is selected If you change that to an unsupported version `External` option automatically gets selected and a warning banner will be displayed that includes a documentation link
- By switching to a different cloud provider or a supported k8s version the warning banner will disappear

https://github.com/user-attachments/assets/dbf8fe0e-9fea-4a55-8f1c-58f4124d62de

## What should be tested

- Watch the video and test the same scenarios
- Also try to finish creating an Azure RKE1 cluster, I did not test that in my video, and there was an issue that got fixed later

### DIFF
--- a/lib/shared/addon/components/cru-cloud-provider/component.js
+++ b/lib/shared/addon/components/cru-cloud-provider/component.js
@@ -183,8 +183,8 @@ export default Component.extend({
       set(configAnswersOut, answer.key, answer.value);
     });
 
-    // Avoid setting cloud provider specific configs if its path doesn't exist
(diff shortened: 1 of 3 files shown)


mantis-toboggan-md (approved):
lgtm

PR #5167, opened 2024-09-05
Remove use of JsZip

nwmac:
This PR removes the JsZip library.

This was only being used in 2 parts of the codebase - cis scans and pipelines - both of these are no longer shown in the UI.

For simplicity, the code has been removed and errors are logged - these code paths should not be reachable in the UI - we will prune this code in the future as part of a larger tidy-up effort.


mantis-toboggan-md (changes requested):

nwmac:
Removed downloadYaml, as the backend API does not work and this is legacy UI

mantis-toboggan-md (approved):

nwmac:
@mantis-toboggan-md Updated following merge of dependabot PR

mantis-toboggan-md (approved):

PR #5178, opened 2024-10-02
update ec2 instances list in amazon.js

tp-drosssmith:
It looks like the included AWS EC2 instance list in Rancher is a bit stale.  I've attempted to update the existing instance types with the latest Intel/AMD versions excluding some of the fancier EC2 types for networking/storage .  Graviton instances look like they're still marked as "experimental" for use in Rancher so those have been excluded too.

### DIFF
--- a/lib/shared/addon/utils/amazon.js
+++ b/lib/shared/addon/utils/amazon.js
@@ -87,6 +87,207 @@ export const INSTANCE_TYPES = [
   },
 
   {
+    group: 'M7i-flex - General Purpose',
+    name:  'm7i-flex.large'
+  },
+  {
+    group: 'M7i-flex - General Purpose',
+    name:  'm7i-flex.xlarge'
+  },
+  {
+    group: 'M7i-flex - General Purpose',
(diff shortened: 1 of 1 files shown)


tp-drosssmith:
This is related to Rancher support case https://scc.suse.com/support/cases/01548430.  My use case is really just to try and save a little money by using AMD instances but once I got started on making the change I saw there was a long list of instance types that probably should be included.

krumware:
+1 here, also need this

mantis-toboggan-md:
It looks like you want to update the list of instances available in the EKS provisioning form specifically - that list should be updated in Rancher 2.9 onwards (we are also updating that data dynamically from an AWS api now so you shouldn't have to keep updating Rancher to get fresh options). 

This PR will however update the list of ec2 options shown in the RKE1 provisioning form so it'd still be nice to get in 2.10. Thanks for submitting this @tp-drosssmith 


Screenshot 2024-10-03 at 9 51 51 AM

mantis-toboggan-md (approved):

PR #5184, opened 2024-11-05
Include ember-api-store addon in the code base

nwmac:
This PR takes the code from https://github.com/rancher/ember-api-store and embeds it in the ui codebase.

`ember-api-store` is currently published as an npm module which complicates Ember migration since we need to migaret both projects. By inlining the code, we only have to migrate the ui codebase.

Note, there is a test directory, but as I can tell, this is all test framework and there are no tests, so this folder was removed.


mantis-toboggan-md:
Is this still needed?

nwmac:
@mantis-toboggan-md not sure it is. This was to help us more easily migrate to newer versions of Ember - but I think we can only go as far as 3.28 - so don’t know if it I worth it.

nwmac:
Won't fix ahead of removal of the legacy UI

PR #5191, opened 2024-11-06
Bump node-sass to 8.0.0

nwmac:
This PR bumps node-saas to 8.0.0 - this updates a number of sub-dependencies, which keeps us up to date.

The latest version is `9.0.0`, but this drops support for Node 14, so we can not use it.

I have checked the change log, and although these are major version bumps, going from 4 to 8, the breaking changes seem small and don't appear to affect things in our usage.

After update, I was able to test the UI and the styling still seemed to be correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "loader.js": "^4.7.0",
     "marked": "^2.0.0",
     "moment": "~2.29.0",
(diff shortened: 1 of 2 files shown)


mantis-toboggan-md (approved):
LGTM - I'm not seeing any style issues either

PR #5222, opened 2025-04-17
Remove legacy unused routes/pages

nwmac:
Addresses https://github.com/rancher/dashboard/issues/12955

Removes some legacy Ember pages no longer needed.

Also removed user preferences from the user menu - the reset password is shown there and not needed, so removed the legacy Ember UI for that.


mantis-toboggan-md (approved):
LGTM

PR #5224, opened 2025-05-07
Add docker install URL for 28.1

eva-vashkevich:
Addresses [rancher/dashboard#14300](https://github.com/rancher/dashboard/issues/14300)

This PR adds the requested docker install URL


nwmac (approved):

nwmac (commented):
@eva-vashkevich I see we are adding both 28.0 and 28.1 - I assume that is correct.

The two PRs are not based off each other, so when they merge, what will be the order of 28.1 and 28.0 in the file they both update?

PR #5230, opened 2025-05-08
Deprecate ember-based UI Plugins

nwmac:
Addresses https://github.com/rancher/dashboard/issues/14286

This PR changes re-uses the RKE1 deprecation banner for deprecation of Ember-based UI plugins.

It also fixes a bug with the code that handles error loading custom javascript for a plugin and now correctly handles this and shows an error on the creation screen when this happens - previously users get a partially-rendered page.

To test:

- Enable a cluster driver from the Drivers page - suggested one is `Open Telekom Cloud CCE`

Go to create and choose this driver.

Check that the page shows the new deprecation banner:

image

Go the Drivers and edit the for the drvier  - add some junk to the end of the URL, eg:

image

Reload the UI (since the plugin is cached and we need to reload to remove it)

Go to the cluster create again and verify that the page shows the new error message and not a partially-rendered form:

image

### DIFF
--- a/app/authenticated/cluster/edit/template.hbs
+++ b/app/authenticated/cluster/edit/template.hbs
@@ -7,7 +7,7 @@
 <BannerMessage
   @color="bg-warning"
   @icon="icon-alert"
-  @message={{t "clustersPage.rke1DeprecationMessage" htmlSafe=true}}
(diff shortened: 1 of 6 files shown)


mantis-toboggan-md (approved):
LGTM

PR #5236, opened 2025-07-09
remove requests to load nodeTemplates during kontainer cluster provisioning

mantis-toboggan-md:
https://github.com/rancher/dashboard/issues/14753

The network tab showed additional 404's, but removing these requests to fetch node templates appears to be sufficient.


app/models/cloudcredential.js
```diff
@@ -9,6 +9,7 @@ const cloudCredential = Resource.extend({
   globalStore:    service(),
   nodeTemplates: hasMany('id', 'nodetemplate', 'cloudCredentialId', 'globalStore'),
 
+
```
eva-vashkevich, line 12: what a power move

eva-vashkevich (approved):
lgtm

